PR #1, opened 2020-04-10
Update to be compatable with gltf-blender-io shipped with blender 2.82

netpro2k:
Updates things to use the new extensibility features added to the gltf exporter that are now shipped in 2.82. This should hopefully make things much more stable since far fewer things are monkeypatched. 

There still remains 1 monkeypatch in place (discussed in https://github.com/KhronosGroup/glTF-Blender-IO/issues/1009). Hopefully we can land something that works for this upstream soon, as this is sure to break on the next update, since a different `gather_gltf_hook` was already introduced upstream (but hasn't landed in the version shipped with Blender yet)

This doesn't add any new features yet but removes the "export scene" and "export selected" buttons, since this is now actually integrated into the normal GLTF export flow.

image


Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+fake-bpy-module-2-82 = "*"
```
robertlong, line 7: Awesome, you figured this out! That'll be a lot better for dev 👍

io_scene_hubs/panels.py
```diff
@@ -46,6 +41,37 @@ class HubsMaterialPanel(Panel):
...
+
+    @classmethod
+    def poll(cls, context):
+        sfile = context.space_data
+        operator = sfile.active_operator
+        return operator.bl_idname == "EXPORT_SCENE_OT_gltf"
+
+    def draw_header(self, context):
+        props = bpy.context.scene.HubsComponentsExtensionProperties
+        self.layout.prop(props, 'enabled', text="")
```
robertlong, line 60: So if you uncheck this, none of the Hubs components stuff runs? Nice!

io_scene_hubs/exporter.py
```diff
@@ -1,282 +0,0 @@
...
-    export_settings = {}
-    export_settings['timestamp'] = datetime.datetime.now()
-    export_settings['gltf_filepath'] = bpy.path.ensure_ext(filepath, filename_ext)
-
-    if os.path.exists(export_settings['gltf_filepath']):
-        os.remove(export_settings['gltf_filepath'])
-
-    export_settings['gltf_filedirectory'] = os.path.dirname(
-        export_settings['gltf_filepath']) + '/'
-    export_settings['gltf_format'] = 'GLB'
```
robertlong, line 200: So, we got rid of these settings, are there any that need to be set for things to work correctly? I remember setting some of these to non-standard values.

io_scene_hubs/panels.py
netpro2k: Yep, its just checked for in each of the hooks

io_scene_hubs/exporter.py
netpro2k: I think we only had to configure these because we were manually calling export. These settings are all set via the checkboxes in the gltf export window.

io_scene_hubs/components.py
```diff
@@ -18,6 +18,18 @@ class HubsComponentName(PropertyGroup):
 class HubsComponentList(PropertyGroup):
     items: CollectionProperty(type=HubsComponentName)
 
+class HubsComponentsExtensionProperties(bpy.types.PropertyGroup):
+    enabled: bpy.props.BoolProperty(
+        name="Export Hubs Components",
+        description='Include this extension in the exported glTF file.',
+        default=True
+        )
+    float_property: bpy.props.FloatProperty(
```
robertlong, line 27: We can remove this

io_scene_hubs/components.py
```diff
@@ -18,6 +18,18 @@ class HubsComponentName(PropertyGroup):
 class HubsComponentList(PropertyGroup):
     items: CollectionProperty(type=HubsComponentName)
 
+class HubsComponentsExtensionProperties(bpy.types.PropertyGroup):
+    enabled: bpy.props.BoolProperty(
+        name="Export Hubs Components",
+        description='Include this extension in the exported glTF file.',
+        default=True
```
robertlong, line 25: I don't know if we should default this to true.

io_scene_hubs/gather_properties.py
```diff
@@ -0,0 +1,73 @@
...
+    elif property_type == 'collections':
+        return gather_collections_property(export_settings, blender_object, target, property_name, property_definition, hubs_config)
+    elif property_type == 'array':
+        return gather_array_property(export_settings, blender_object, target, property_name, property_definition, hubs_config)
+    else:
+        return gltf2_blender_extras.__to_json_compatible(getattr(target, property_name))
+
+def gather_array_property(export_settings, blender_object, target, property_name, property_definition, hubs_config):
+    array_type = property_definition['arrayType']
+    print(array_type)
```
robertlong, line 30: You missed a print()

io_scene_hubs/__init__.py
```diff
@@ -1,30 +1,123 @@
...
+from .gather_properties import gather_properties
 
 bl_info = {
     "name" : "io_scene_hubs",
     "author" : "Robert Long",
     "description" : "",
-    "blender" : (2, 80, 0),
-    "version" : (0, 0, 1),
+    "blender" : (2, 82, 0),
+    "version" : (0, 0, 2),
```
robertlong, line 14: We should probably bake the exporter extension version into the gltf file. We can either append it to the `gltf.asset.generator` string or put it in `gltf.asset.extras.HUBS_gltfExporterVersion` or something like that.

fernandojsg: yeah that would be very useful

fernandojsg:
> Just a few nits but it looks great! Also, if you could help me understand pipenv that'd be awesome. I just made my python environment on my mac a disaster 😅

+1 for a quick course on pipenv :_D

antpb:
I tested this out with Blender 2.82 (using [gltf exporter's 2.83 branch](https://github.com/KhronosGroup/glTF-Blender-IO/tree/blender-v2.83-release)) and things seems to be working well! I am able to add a scale-audio-feedback component to an object and export it out for Spoke. In the Spoke gltf validator I'm seeing a warning about MOZ_hubs_components so that seems to indicate that this tool is adding the components to my final GLTF. Not yet seeing the scaling behavior, but I know that the components are definitely being added.

Without using the GLTF exporter's 2.83 branch I _do not_ get the panel for Hubs Components like in the screenshot of this PR description. Maybe that is a bug? The `Extensions` menu option does show, but when you expand, the Hubs options are not populating. It's just blank. Hope this testing helps.

io_scene_hubs/components.py
netpro2k: It will only do anything if you actually have hubs components added, in which case I think it makes sense to include them. Easy enough to "save export settings" and turn it off if that's not the case for some reason.

netpro2k:
> I tested this out with Blender 2.82 (using [gltf exporter's 2.83 branch](https://github.com/KhronosGroup/glTF-Blender-IO/tree/blender-v2.83-release)) and things seems to be working well! I am able to add a scale-audio-feedback component to an object and export it out for Spoke. In the Spoke gltf validator I'm seeing a warning about MOZ_hubs_components so that seems to indicate that this tool is adding the components to my final GLTF. Not yet seeing the scaling behavior, but I know that the components are definitely being added.
> 
> Without using the GLTF exporter's 2.83 branch I _do not_ get the panel for Hubs Components like in the screenshot of this PR description. Maybe that is a bug? The `Extensions` menu option does show, but when you expand, the Hubs options are not populating. It's just blank. Hope this testing helps.

Hmm, I have only tested this against the version of the GLTF exporter that actually is bundled with Blender 2.82, given the nature of some of the monkeypatching we are still having to do, I think supporting multiple versions of the gltf exporter is going to be problematic. This should get much better over time as we can land these changes upstream and not have to rely on brittle patching (this has already gotten WAY WAY better in this branch)

PR #3, opened 2020-04-16
Rename references of Hubs to Moz Components

fernandojsg:
I was giving a try to rename the `Hubs` references with a more generic Moz Components as we discussed.
I used the reference branch from @netpro2k #1 
Do you have any strong preference if, when using variables like `hubs_settings` replace them to `mozcomponents_settings` or just `moz_settings` ?
I tested it already in blender and it works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,3 @@
-# Hubs Exporter for Blender
+# GLTF MOZ_Components Exporter for Blender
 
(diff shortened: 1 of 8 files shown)


robertlong:
I think we should close this since we aren't moving to the `MOZ_components` extension name soon.

PR #4, opened 2020-06-03
Add lightmap support and upgrade to Blender 2.83

netpro2k:
This adds support for `MOZ_lightmap` exporting and upgrades to be compatible with Blender 2.83 (it is **NOT** backwards compatible with 2.82, or more specifically the GLTF addon that ships with 2.82)

To export a lightmap, create a "Group" node named `MOZ_lightmap` with a color input property named `Lightmap` and a float input property named `Intensity`. Use a `UV Map` node to control what UV set should be used for the lightmap, as you would any other texture in Blender. 

image

Note that fore use in Hubs, you currently MUST use the SECOND UV set, as ThreeJS is currently hardcoded to use that for lightmaps. This will likely be fixed in the future so the exporter does not enforce this.

image

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -83,7 +83,12 @@ class glTF2ExportUserExtension:
         # We need to wait until we create the gltf2UserExtension to import the gltf2 modules
         # Otherwise, it may fail because the gltf2 may not be loaded yet
         from io_scene_gltf2.io.com.gltf2_io_extensions import Extension
+        from io_scene_gltf2.blender.exp import gltf2_blender_get
+        from io_scene_gltf2.blender.exp import gltf2_blender_gather_texture_info
+
         self.Extension = Extension
+        self.gltf2_blender_get = gltf2_blender_get
+        self.gltf2_blender_gather_texture_info = gltf2_blender_gather_texture_info
         self.properties = bpy.context.scene.HubsComponentsExtensionProperties
         self.hubs_settings = bpy.context.scene.hubs_settings
         self.was_used = False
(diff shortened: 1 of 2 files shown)


__init__.py
```diff
@@ -23,27 +23,18 @@
...
 from io_scene_gltf2.blender.exp import gltf2_blender_export
 from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 orig_gather_gltf = gltf2_blender_export.__gather_gltf
 def patched_gather_gltf(exporter, export_settings):
     orig_gather_gltf(exporter, export_settings)
-    export_user_extensions('gather_gltf_hook', export_settings, exporter._GlTF2Exporter__gltf)
+    export_user_extensions('hubs_gather_gltf_hook', export_settings, exporter._GlTF2Exporter__gltf)
     exporter._GlTF2Exporter__traverse(exporter._GlTF2Exporter__gltf.extensions)
 
-# Monkey patch to add gather_joint_hook, must be removed once handled upstream
```
netpro2k, line 35: This is now merged upstream https://github.com/KhronosGroup/glTF-Blender-IO/pull/1021

__init__.py
```diff
@@ -23,27 +23,18 @@
...
-# Monkey patch to add gather_gltf_hook, must be removed once handled upstream https://github.com/KhronosGroup/glTF-Blender-IO/issues/1009
+# gather_gltf_hook does not expose the info we need, make a custom hook for now
+# ideally we can resolve this upstream somehow https://github.com/KhronosGroup/glTF-Blender-IO/issues/1009
 from io_scene_gltf2.blender.exp import gltf2_blender_export
 from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 orig_gather_gltf = gltf2_blender_export.__gather_gltf
 def patched_gather_gltf(exporter, export_settings):
     orig_gather_gltf(exporter, export_settings)
-    export_user_extensions('gather_gltf_hook', export_settings, exporter._GlTF2Exporter__gltf)
+    export_user_extensions('hubs_gather_gltf_hook', export_settings, exporter._GlTF2Exporter__gltf)
```
netpro2k, line 33: A `gather_gltf_hook` was added upstream https://github.com/KhronosGroup/glTF-Blender-IO/commit/0827c6d857d2156b09375eef9cefddf5924fb5c8#diff-37cfcca6b43542789078e990ff811ea4 but it does not expose the attributes we actually need, so continue to apply a monkey patch. Discussion about this happening here https://github.com/KhronosGroup/glTF-Blender-IO/issues/1009

fernandojsg: fwiw if we will be moving towards `s/hubs/moz`(as on #3 ) could it be better to use `moz_gather_gltf_hook`?

fernandojsg:
So regarding backward compatibility, are we just assuming that we will be compatible just with the latest version right?

__init__.py
netpro2k: Yeah I figure it would be best to just do that all at once in #3 so we can see it all in one spot. I think I will tackle that alongside reworking how settings are configured.

netpro2k:
> So regarding backward compatibility, are we just assuming that we will be compatible just with the latest version right?
Yeah right now I am just taking the stance that each version of this addon will only work with the latest blender when that addon version was released. If this causes people significant headache we can look into supporting multiple versions, but right now because of how its integrated with the GLTF exporter its going to break fairly often as they release updates. This should get more stable over time as we have to monkeypatch less and use fewer internal APIs.

PR #6, opened 2020-06-09
Improvments to components panel

netpro2k:
A bunch of miscellaneous cleanups and improvements to component interface
image

- [x] cleanup components UI + make component properties collapsible
- [x] add `nav-mesh`, `uv-scroll`, and `waypoint` as default components
- [x] fix exporting vec properties
- [x] add support for enum properties
- [x] add support for image properties
- [x] support for description, subType, and unit annotation on properties
- [x] support for grouping components into categories
- [x] fix python warnings about "attributes" usage on dynamic classes
- [x] ~~move component config to addon settings instead of scene node~~ Moved to its own PR #7

### DIFF
--- a/components.py
+++ b/components.py
@@ -58,15 +58,16 @@ def define_class(class_name, class_definition, hubs_context):
     if class_name in registered_classes:
         return registered_classes[class_name]
 
-    class_property_dict = {}
+    class_property_dict = {
+        '__annotations__': {},
+        'definition': class_definition
+    }
 
     for property_name, property_definition in class_definition['properties'].items():
         property_class = define_property(class_name, property_name, property_definition, hubs_context)
 
         if property_class:
-            class_property_dict[property_name] = property_class
-
-    class_property_dict['definition'] = class_definition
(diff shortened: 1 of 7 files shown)


components.py
```diff
@@ -100,61 +102,102 @@ def define_property(class_name, property_name, property_definition, hubs_context
 
     if property_type == 'int':
         return IntProperty(
-            name=property_name
+            name=property_name,
+            description=property_definition.get("description") or "",
+            subtype=property_definition.get("subType") or "NONE",
```
robertlong, line 107: This looks interesting, what can we use subtypes on for these properties?

netpro2k: There are a bunch of different sub types like ‘VELOCITY’,, ‘EULER’, ‘QUATERNION’, ‘AXISANGLE’, ‘XYZ’ that just show slightly nicer UI. There are also units like 'LENGTH", 'MASS', 'TIME', etc which will show the unit in the field. I don't know that these need to be part of any official component "schema' we come up with, but seems nice to expose as optional since it doesn't actually impact the data itself, just the presentation.

PR #7, opened 2020-06-12
Refactor settings

netpro2k:
This moves addon configuration off of the scene object and into the preferences panel. This also makes these settings non file specific and cleans up some of the hoops we had to jump through. 

image

Also centralizes the way we access settings, previously we sometimes would do it through a global context object and sometimes through the scene node, it is now always done through `hubs_context` in the settings module. I think it can be further cleaned up, but makes sense to do in another pass I think since this already moved enough around. Once that is done I think we can tackle #3.

- [x] point this PR at master once the base branch has been merged

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -4,9 +4,11 @@ import bpy
 import pathlib
 
 from bpy.props import StringProperty, PointerProperty
-from bpy.types import PropertyGroup
-from bpy.app.handlers import persistent
+from bpy.types import PropertyGroup, AddonPreferences
+import pathlib
+import os
 from . import components
+from .operators import ReloadHubsConfig
 
 # Get the path to the default config file
 main_dir = pathlib.Path(os.path.dirname(__file__)).resolve()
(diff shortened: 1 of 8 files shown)


settings.py
```diff
@@ -111,49 +112,32 @@ def register_components():
...
-
-    def reload_config(self):
-        reload_context(self.config_path)
+    def draw(self, context):
+        layout = self.layout.row()
+        layout.prop(self, "config_path")
+        layout.operator(ReloadHubsConfig.bl_idname, text="", icon="FILE_REFRESH")
 
 def register():
-    bpy.utils.register_class(HubsSettings)
```
robertlong, line 149: I'm a little worried about how this will impact existing component data, but I don't have a good project to test with. It could be mostly fine. You may just need to update the path to the Hubs components config.

robertlong (commented):
LGTM, I resolved the merge conflicts and did some brief testing locally. Having the config in the addon preferences section seems much better.

keianhzo:
This is not required anymore after the new architecture update

PR #11, opened 2020-08-05
Media components

rawnsley:
I've added some of the common component types for export: visible, link, image, audio, and video. The media components all require a *networked* component with a UUID, which allows NAF to track changes. This is added dynamically during export with a new UUID each time as there's no reason that I could think of for this to be persistent.

This pull request includes the previous *visibility* component change that is essential for tidy *nav-mesh* export and desirable for some uses of the *audio* component.

The driver for this is the desire to allow designers to export scenes directly from Blender without needing to go via Spoke. Scene files can be uploaded directly to a Hubs room with the "Custom Scene" option. This greatly simplifies our workflow when dealing with large numbers of scenes and speeds up iteration time.

Feel free to step all over the implementation. I'm an enthusiastic amateur on Blender and an unenthusiastic amateur on Python so it may not be very idiomatic.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -1,4 +1,5 @@
 import bpy
+import uuid
 
 from . import settings
 from . import components
@@ -148,6 +149,7 @@ class glTF2ExportUserExtension:
 
         if component_list.items:
             extension_name = hubs_config["gltfExtensionName"]
+            is_networked = False
             component_data = {}
 
             for component_item in component_list.items:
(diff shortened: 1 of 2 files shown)


netpro2k:
Excellent work! I very much echo the sentiment "'I'm an enthusiastic amateur on Blender and an unenthusiastic amateur on Python", its very likely most of the extension code is not idiomatic Python or Blender code, but it seems to be getting the job done. I had actually started implementing the network stuff as well but was making it a separate component. Doing it this way seems better as it better mirrors what happens in spoke. Eventually I think Hub's should be the one taking care of this since the "networked" component is an implementation detail and should probably not actually be in the gltf... 

Sorry it took so long to get this merged, but its going into the next release.

PR #12, opened 2020-08-05
Nav mesh property hack

rawnsley:
Potential workaround for [the nav-mesh export problem](https://github.com/MozillaReality/hubs-blender-exporter/issues/10).

Apologies for including the visibility change in here, but I've only just remembered that it's tidier if each pull request is on its own branch from the upstream master.

### DIFF
--- a/default-config.json
+++ b/default-config.json
@@ -39,7 +39,7 @@
       "category": "Scene",
       "node": true,
       "properties": {
-      "visible": { "type": "bool", "default": true }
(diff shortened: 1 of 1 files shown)


netpro2k:
I think I am going to just implement this fix at the serialization layer so that any empty components just automatically work and you don't have to muddy the component schemas fixing it. Thanks for spelling out the core issue and fix though, made it real easy to implement a fix.

PR #14, opened 2020-09-17
Root scene components: background, fog, and audio-settings

rawnsley:
All of the components that can be added to the root of a Hubs scene. I added them to a new category called "Root" to differentiate them from components that can be added to arbitrary nodes, but I couldn't see an easy way to enforce this restriction.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -1,4 +1,5 @@
 import bpy
+import uuid
 
 from . import settings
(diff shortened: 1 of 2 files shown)


rawnsley:
I've withdrawn this for now as there are problems I hadn't noticed. Apologies for the confusion.

PR #15, opened 2020-09-18
Scene-level components

rawnsley:
I've added support for the three components that can be defined at the scene level in Spoke:

- Background color
- Fog settings
- Default audio settings (avatars and media)

Colors in Blender appear to be defined in linear space and the raw linear values are shown in the RGB part of the color editor. However the hex values in the color editor are in a gamma-corrected space. I have assumed that the color fields for the background and fog Hubs components should be linear colors (I think this is typical of single value GLTF colors) and so the slightly unexpected consequence is that if you paste a hex color into Blender, you won't necessary get the same hex codes out again in the GLTF file. This might need some scrutiny as my understanding of color spaces is limited.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -112,6 +112,16 @@ class glTF2ExportUserExtension:
             gltf2_object.asset.extras = {}
         gltf2_object.asset.extras["HUBS_blenderExporterVersion"] = get_version_string()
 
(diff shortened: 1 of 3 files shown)


netpro2k:
These look good to me. I am not actually too sure on the color space issues myself, but it sounds like what you have done should match up runtime perceived color, which seems reasonable to me. We can revisit later if it turns out people want more flexibility on this.

PR #20, opened 2020-12-23
add installation instructions

### LINKED ISSUE #16: Install instructions
It would be helpful to have installation instructions in the readme so that folks know to download the zip file, click install in the preferences, and choose the zip file from the file browser.  This was only the second addon I installed in Blender and it's been a while since I installed the first one, so I had forgotten what to do and it took a while to figure it out.

matthewbcool:
closes #16


netpro2k:
Thanks for submitting this!

PR #30, opened 2021-04-28
Image property "controls" isn't required or used anywhere

rawnsley:
I can't see any code paths in Hubs client where this is actually used nor can I think of a scenario where it might be. The same property is surfaced in Spoke, but doesn't appear to do anything there either.

Happy to be wrong, but otherwise I suggest the property is removed to avoid future confusion.

### DIFF
--- a/default-config.json
+++ b/default-config.json
@@ -188,11 +188,6 @@
           "type": "string",
           "description": "Image URL"
         },
-        "controls": {
-          "type": "bool", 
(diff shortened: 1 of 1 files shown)


rawnsley:
Ah, my original testing was incomplete. Setting _controls_ to true for an image enables the **Inspect** and **Refresh** buttons, both of which have value. Closing the PR as invalid.

PR #32, opened 2021-05-14
Loading models from a URL

rawnsley:
The addition of a `model` component with a _src_ property that will load a model in the same way image, video, and audio components do.

The only change that is required on the Hubs side is the correction of [a bug in the gltf-component-mapping code](https://github.com/mozilla/hubs/blob/master/src/gltf-component-mappings.js#L274) where both _fitToBox_ and _moveTheParentNotTheMesh_ are true, but this is prohibited later in media-loader. This change will be submitted as part of a future PR on the Hubs codebase.


netpro2k:
Rebased and merged this in manually https://github.com/MozillaReality/hubs-blender-exporter/commit/3b6eeaca3b7615c6813ee7c80e22577ec18f863d

Thanks!

PR #33, opened 2021-05-14
Moveable media component

rawnsley:
This PR is a companion to [this PR](https://github.com/mozilla/hubs/pull/4261) on the Hubs codebase.

The `moveable` component allow media elements to be "blessed" with the ability to be moved around in the scene as if they were added dynamically to the room at runtime. This opens up a number of valuable use cases.

The component can be applied to any node, not just those with media components, and that will allow part of the scene to be moved independently. An example might a model of an engine where nodes in the scene graph were made "moveable" to allow disassembly.


keianhzo:
Closing as the client counterpart has been also closed.

PR #36, opened 2021-07-10
Add audio zones support and refactor to audio-params

keianhzo:
This PR adds support for the audio zone component in Blender.

It also refactors the audio parameters from `audio`, `video` and `audio-target` components to use the `audio-params` component.

Related https://github.com/mozilla/hubs/pull/4399
Related: https://github.com/mozilla/Spoke/pull/1156

### DIFF
--- a/components.py
+++ b/components.py
@@ -307,6 +307,10 @@ def add_component(obj, component_name, hubs_config, registered_hubs_components):
             else:
                 component[property_name] = default_value
 
+    if 'deps' in component_definition:
+        for dep in component_definition["deps"]:
+            add_component(obj, dep, hubs_config, registered_hubs_components)
+
 def remove_component(obj, component_name):
     items = obj.hubs_component_list.items
     items.remove(items.find(component_name))
--- a/components.py
+++ b/components.py
@@ -279,6 +279,11 @@ def get_wildcard(arr, path_parts):
     return values
 
 def add_component(obj, component_name, hubs_config, registered_hubs_components):
+    if any(item.name == component_name for item in obj.hubs_component_list.items):
+        print('Hubs component \'%s\' already exists' % (
+            component_name))
+        return
+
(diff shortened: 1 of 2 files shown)


components.py
```diff
@@ -307,6 +307,10 @@ def add_component(obj, component_name, hubs_config, registered_hubs_components):
             else:
                 component[property_name] = default_value
 
+    if 'deps' in component_definition:
+        for dep in component_definition["deps"]:
+            add_component(obj, dep, hubs_config, registered_hubs_components)
```
netpro2k, line 312: This doesn't check if the object already has this component.. It might make sense to just do that in add_component itself. Or maybe both (with add_component logging a warning)

default-config.json
```diff
@@ -310,62 +310,9 @@
           "type": "bool", 
           "description": "Loop",
           "default": true
-        },
-        "audioType": {
```
netpro2k, line 314: Slightly concerned people will lose their settings for these when upgrading to this version of the addon.. but not sure its worth writing special code to migrate this.. hmm

default-config.json
```diff
@@ -934,34 +852,26 @@
...
           "default": 0.0
-        },
-        "debug": {
-          "description": "Show debug visuals.",
-          "type": "bool",
-          "default": false
         }
       }
     },
-    "zone-audio-source": {
```
netpro2k, line 945: Sidenote, we may need to rethink the name of this...

netpro2k (commented):
This mostly lgtm. Made a minor note, but mostly am wondering what should be our policy on migration of breaking changes like this.

components.py
keianhzo: I've added a check to avoid duplicated components

default-config.json
keianhzo: I think we agreed on leaving it until we merger a few components together so leaving like this for the moment.

PR #39, opened 2021-09-10
Add "shadow" component

rawnsley:
The "shadow" component isn't available for Blender projects, which limits complex scene lighting.


netpro2k:
Doing a new release so going through the PR backlog, thanks for adding this!

PR #40, opened 2021-09-16
Added missing spot-light component

rawnsley:
This spot-light component completes the set of supported light types. I included a min-max range to match the Spoke defaults.

The min-max implementation takes the default values (+/- 3.402823e38) from [the Blender documentation](https://docs.blender.org/api/current/bpy.props.html#bpy.props.FloatProperty) and approximates them as +/- 1e38 to avoid any number parsing issues. If I was better at Python I could probably express this more elegantly.

Related: the "Elements" group is growing pretty big and we might want to consider putting the lights into a group of their own along with **shadow** and possibly **visible**.

### DIFF
--- a/components.py
+++ b/components.py
@@ -121,7 +121,9 @@ def define_property(class_name, property_name, property_definition, hubs_context
             name=display_name,
             description=property_definition.get("description") or "",
             subtype=property_definition.get("subType") or "NONE",
-            unit=property_definition.get("unit") or "NONE"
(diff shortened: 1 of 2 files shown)


netpro2k:
Confusing that the docs for FloatProperty don't mention min/max (then mention soft and hard min/max) but examples seem to use it.. Wonder if we should use `float('inf') as the default instead

PR #44, opened 2021-10-04
OpenEXR export

rawnsley:
Added support for OpenEXR by defining a new extension `MOZ_texture_exr`

Here's an example Blender scene with EXR images for testing: https://github.com/LearnHub/HubHenge/tree/exr-version


rawnsley:
@keianhzo coincidentally this PR addresses [this comment](https://github.com/MozillaReality/hubs-blender-exporter/blob/2e9d0b841a39c1647ed2c3ec39344a595ee2523e/addons/io_hubs_addon/io/utils.py#L42) you added last year

keianhzo (approved):
Great, thanks for the contribution!

Exairnous:
The code seems good.  This isn't a user facing change, so I don't see any problem with adding it, but I thought that supporting EXR wasn't a direction the Hubs team wanted to go in right now, so has that changed, or is this something you need for your client?  I think EXR support is nice to have, so thanks :) I'm a bit curious though.

rawnsley:
I discussed our use case [in this companion PR](https://github.com/mozilla/hubs/pull/4706), but basically EXR format is a better intermediate format to go into our post-processing tools as it preserves precision and has a wide range of layout options including alpha channels and file compression. We added support in Hubs itself because it is sometimes useful to review the scene without post-processing and compare before and after.

Exairnous:
Ah, okay.  Thank you.

PR #58, opened 2021-12-29
Frustrum culling component

rawnsley:
Companion PR for https://github.com/mozilla/hubs/pull/4804

Simple boolean component to disable frustrum culling. Only expected to be required for animations with extreme changes of vertex position due to skinning.


keianhzo (approved):
LGTM

PR #63, opened 2022-01-25
Add support for importing Hubs components from glTF files

Exairnous:
This PR adds support for importing Hubs components from glTF files.  All components from objects, materials, and the scene should be imported correctly.

This PR operates by hooking into the default glTF importer's create object and material methods to first gather a list of glTF nodes with Hubs components and then hooks into the glTF importer's create scene method to add all of the components in their relative places and apply their settings.  The data parsing is fairly robust and handles most parameters by default, so it should handle most new components added in the future automatically, as long as they don't require special handling (such as linking to objects like the Video Texture Target component does).  If an error occurs while adding a component, it is printed to the terminal and the import continues on.  Component information is also currently printed to the terminal.


Exairnous:
Here is the GLB file I've been testing importing with:
[import_dev_test_6.zip](https://github.com/MozillaReality/hubs-blender-exporter/files/7947830/import_dev_test_6.zip)

Exairnous:
Mentioning @netpro2k and @keianhzo since they've shown some interest in this topic previously. :)

Exairnous:
This PR doesn't properly support files exported from Spoke.  I will attempt to rectify the problem.

Exairnous:
Updated to the new add-on architecture.  Support for lightmaps, and probably bones, still needs to be added, and a bug with the text component needs to be fixed.

keianhzo (commented):
Sorry for the delay reviewing this and thanks a lot for the contribution, this is quite an essential part of the future of the add-on.

I've been testing this and it's a great starting point. As there were quite a few things that I wanted to draft/propose I've created a PR in your fork to avoid making this PR too messy.

I think that although we need to support LTS, we should also support the new public official API for a smoother transition and to avoid breakage (although it also can happen with the public one in early stages). There is not much more code necessary for supporting both so I've added support in a consolidated importer.

I've also moved all the component specific logic to the components similarly to how we handle the component specific export logic. This makes the importer/exporters much leaner and readable.

Other that that, there are a few things that need to be addressed, you pointed a few and I'm adding a few more:
- [ ] Support for lightmaps
- [ ] Support for bones
- [ ] Fix a bug with the text component
- [ ] Loop animation actions are not imported correctly
- [ ] Add tests

Exairnous:
> Sorry for the delay reviewing this and thanks a lot for the contribution, this is quite an essential part of the future of the add-on.

That's okay, we've both been busy with other aspects of the add-on.

> I've been testing this and it's a great starting point. As there were quite a few things that I wanted to draft/propose I've created a PR in your fork to avoid making this PR too messy: [Exairnous#1](https://github.com/Exairnous/hubs-blender-exporter/pull/1)

Sounds good.

> I think that although we need to support LTS, we should also support the new public official API for a smoother transition and to avoid breakage (although it also can happen with the public one in early stages). There is not much more code necessary for supporting both so I've added support in a consolidated importer.

This seems reasonable, and if there are any problems with the new import hooks we have something to fall back on.

> I've also moved all the component specific logic to the components similarly to how we handle the component specific export logic. This makes the importer/exporters much leaner and readable. There is now a new `gather_import` class method to handle component creation in the blender object. The base method applies the generic behavior and a few components override it for specific cases.

Interesting, that definitely makes it more consistent.

> Other that that, there are a few things that need to be addressed, you pointed a few and I'm adding a few more:
> 
>     * [ ]  Support for lightmaps
> 
>     * [ ]  Support for bones
> 
>     * [ ]  Fix a bug with the text component
> 
>     * [ ]  Loop animation actions are not imported correctly
> 
>     * [ ]  Add tests
> 
> 
> My idea is merging my proposed PR into this branch (or cherry pick and update parts) so we can continue working on this PR once the branch has been updated. If you have a better idea of how to handle it let me know.

Thank you for the updated bug list.  I think that sounds like a good plan.  Now that things are stabilizing with the rest of the add-on, I should be able to focus on the importer more.

Exairnous:
> @Exairnous what's exactly the text component issue that you mentioned?

I encountered it when importing a glb exported from Spoke that contained a text component.  It seems to still be present in your branch.

Here is the traceback that my branch printed:
```
Traceback (most recent call last):
  File "/home/ryan/.config/blender/3.1/scripts/addons/io_hubs_addon/io/gltf_importer.py", line 97, in create_object_hubs_components
    assign_property(gltf, blender_component, property_name, property_value)
  File "/home/ryan/.config/blender/3.1/scripts/addons/io_hubs_addon/io/gltf_importer.py", line 331, in assign_property
    setattr(blender_component, property_name, property_value)
TypeError: bpy_struct: item.attr = val: Text.outlineBlur expected a string type, not int
```

keianhzo:
@Exairnous I've fixed some conflicts. I'm going to look into fixing the pending things, I'd like to get users testing this soonish.

Exairnous:
Thanks.  Yeah, it's getting close to where user testing will be helpful.

keianhzo:
This looks great, I think we are very close to having this working. The tests pass which is a great sign. The only issue that I see so far is the linter issues. What else is left?

Exairnous:
Thanks!  Yeah, it's pretty close, I just have to remove the reflection probe changes (as discussed), make the box collider component convert the object scale which Spoke uses to determine the area to the half extents used by the ammo shape, add a warning for non [1,1,1] scales to the import report for ammo shapes/media frames, and do a final format.

Then I'm going to open separate PRs to switch the override on audio-params to false (this is required for correct imports), add the mirror component, and add back the fog color property that was accidentally removed in a recent PR.

I hope to have this all done by the end of my day.

Exairnous:
I ran into a couple of problems I didn't account for and had to solve (I did get them solved, though), so I'm not going to make it by the end of my day, but I should make it by the end of my next day.

Exairnous:
@keianhzo Okay, so I think this is ready for review now.  The heightfield component isn't imported, but I believe it isn't needed for Blender equivalent so I've set it to be ignored.  The trigger volume component isn't imported either, but that gives the user a warning.  Maybe in the future, trigger volume components could be mapped to behavior graphs.

The box collider required special support to import properly.  I think everything is working properly now, but maybe give it an extra once over just to make sure.

I've opened the other PRs for the mirror and audio override so they can be imported properly as well.

The one case that doesn't import is if there is an audio-params component by itself on an object (without any audio or video components), but that's not a valid setup and I think it's highly unlikely to occur in a real use case.

keianhzo:
Awesome, thanks for working on this. I'll review on Monday.

> @keianhzo Okay, so I think this is ready for review now. The heightfield component isn't imported, but I believe it isn't needed for Blender equivalent so I've set it to be ignored. The trigger volume component isn't imported either, but that gives the user a warning. Maybe in the future, trigger volume components could be mapped to behavior graphs.

There are a few components in Spoke that we don't support but I don't expect many people going from Spoke to Hubs so I'm not super concerned about not supporting those. 

> The one case that doesn't import is if there is an audio-params component by itself on an object (without any audio or video components), but that's not a valid setup and I think it's highly unlikely to occur in a real use case.

Audio zones can have `audio-params` without autio/video components, those should still import.

Exairnous:
> Awesome, thanks for working on this. I'll review on Monday.

You're welcome.  Sounds good.

> There are a few components in Spoke that we don't support but I don't expect many people going from Spoke to Hubs so I'm not super concerned about not supporting those.

Honestly, I still think importing from Spoke is going to be the main use case for this, especially with Mozilla ceasing development/hosting of Hubs and people wanting to back up all their stuff.  I thought I had accounted for all the Spoke components except for the heightfield and trimesh (which I think I recall asking about and getting the answer that they weren't needed for the Blender process), and the trigger volume component (which was experimental and not advised to be used anyway).  If there are any others that aren't being supported, or if I'm mistaken in any of this, let me know and I'll add support for whatever is missing.

> Audio zones can have audio-params without autio/video components, those should still import.

Ah, sorry. What I probably should have said there is "without any audio or video _related_ components". Anything that automatically adds/removes the audio params component in Blender, is fully supported for import, but in previous versions of the exporter you were able to add an audio-params component by itself without anything else, so that is the one case I was referring to that I didn't support (and I don't think needs to be supported because having a singular audio-params component on an object without anything else is an invalid setup).

Exairnous:
Hey, so I've added in support for the Clip Rect property on text components.  I had initially planned to do this as a separate PR, but since it needs a custom gather_import I figured it would be easier to just add it here.  Also, do you know if it's valid for the Clip Rect Min values to go above zero, or for the Clip Rect Max values to go below zero? (I'm wondering if we should enforce hard limits there for the exporter even though Spoke doesn't).

keianhzo:
Merged, thanks!

PR #71, opened 2022-06-09
Blender 3.2.0 updates

keianhzo:
Exporter updates for 3.2.0 compat


keianhzo:
The Beta has landed and it already includes this update

PR #72, opened 2022-06-28
Add bone migration.

Exairnous:
Add migration support for bones.  Note: with this migration, media frames that are on bones will draw their gizmos incorrectly (Y and Z will be inverted) in Blender, but it exports to Hubs correctly so it seems to be the gizmo drawing code that isn't accounting for bones properly.


keianhzo (approved):
LGTM. Thanks for this.

PR #82, opened 2022-07-13
Hubs Add Component Menu Improvement Experiment

Exairnous:
Sort categories and components alphabetically, and allow for multiple rows in the menu,
which is user configurable via a row length property in the add-on's new preferences.

Also, separated out some of the components in the Elements category into
their own categories: A/V Elements, and Lights.

Note: the logic seems to be pretty solid, but it can be a bit confusing.  Splitting the sorting and organizing code off into its own function (or multiple functions) may be desirable.

### DIFF
--- a/addons/io_hubs_addon/__init__.py
+++ b/addons/io_hubs_addon/__init__.py
@@ -1,5 +1,6 @@
 from .io import gltf_exporter
 from . import (nodes, components)
+from . import preferences
 bl_info = {
     "name": "Hubs Blender Addon",
     "author": "Mozilla Hubs",
@@ -16,6 +17,7 @@ bl_info = {
 
 
 def register():
+    preferences.register()
     gltf_exporter.register()
     nodes.register()
     components.register()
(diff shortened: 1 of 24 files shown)


netpro2k:
This looks great!

I agree that I would change "AV Elements" to "Media". I would also move "Model" into there (and maybe Link and Media Frame too? though less sure about that).

Also I would move "Frustum", "Billboard", "Shadow", and "Ammo Shape" into "Object" (since they are all modifying object properties) and move "Waypoint" into "Elements".

I like the idea of starting to build out some preferences too, Blender is all about making the exact editor for your specific workflow.

Exairnous:
Sure, "Media" is good, I'll do that.  I'll also update it with the category changes Dom has suggested, but I'm not sure "Model" really belongs in the Media category.  I'll move it for now and we can see what people think.  Link and Media Frame I would keep where they are because Link isn't necessarily media related and Media Frame is a Hubs specific element.

keianhzo (approved):
This looks good to me for now. We can address the frustum typo in another PR. Thanks for this!

PR #88, opened 2022-07-19
Move reflection probe resolution and baking to a global scene prop

keianhzo:
This PR moves the reflection probes environment resolution and bake button to the scene panel as global properties. Now all reflection probes will use the same env map resolution. Also:

- Adds support for multiple reflection probe env map renders
- Improves error handling and logging during baking reporting to the info area
- Saves/restores previous render settings after the baking finishes
- Adds a preference to customize a global tmp path to store all addon generated temporary files
- Adds a new hubs component method to allow components to draw global properties in any panel (for now it mostly makes sense in the scene one). The global properties are drawn before the component blocks.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/reflection_probe.py
+++ b/addons/io_hubs_addon/components/definitions/reflection_probe.py
@@ -301,10 +301,7 @@ class ReflectionProbe(HubsComponent):
         row = layout.row()
         row.label(text="You can bake all reflection probes at once from the scene settings.",
                   icon='INFO')
-        layout.separator()
         super().draw(context, layout, panel_type)
-
-        layout.separator()
 
         bake_msg = "Baking..." if probe_baking else "Bake"
         bake_op = layout.operator(
@@ -312,6 +309,12 @@ class ReflectionProbe(HubsComponent):
             text=bake_msg
         )
         bake_op.bake_selected = True
+
+        if not hasattr(bpy.context.scene, "cycles"):
+            row = layout.row()
+            row.alert = True
+            row.label(text="Baking requires Cycles addon to be enabled.",
+                      icon='ERROR')
 
     def gather(self, export_settings, object):
         return {
@@ -340,18 +343,18 @@ class ReflectionProbe(HubsComponent):
                 row.label(text="Reflection probe resolution has changed. Bake again to apply the new resolution.",
                           icon='ERROR')
 
+            bake_msg = "Baking..." if probe_baking else "Bake All"
+            bake_op = col.operator(
+                "render.hubs_render_reflection_probe",
+                text=bake_msg
+            )
+            bake_op.bake_selected = False
+
             if not hasattr(bpy.context.scene, "cycles"):
                 row = col.row()
                 row.alert = True
                 row.label(text="Baking requires Cycles addon to be enabled.",
                           icon='ERROR')
-
-            bake_msg = "Baking..." if probe_baking else "Bake All"
-            bake_op = layout.operator(
-                "render.hubs_render_reflection_probe",
-                text=bake_msg
-            )
-            bake_op.bake_selected = False
 
     @classmethod
     def poll(cls, context, panel_type):
--- a/addons/io_hubs_addon/components/definitions/reflection_probe.py
+++ b/addons/io_hubs_addon/components/definitions/reflection_probe.py
@@ -33,6 +33,7 @@ RESOLUTION_ITEMS = [
 ]
 
 probe_baking = False
+show_warning = False
 
 
 def get_probes():
@@ -242,6 +243,10 @@ class ReflectionProbe(HubsComponent):
     )
 
     def draw(self, context, layout, panel_type):
+        if show_warning:
+            row = layout.row()
+            row.label(text="Reflection probes resolution is now set globally in scene settings.",
+                      icon='INFO')
         super().draw(context, layout, panel_type)
 
     def gather(self, export_settings, object):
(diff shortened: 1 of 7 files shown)


Exairnous:
Hi, some nice improvements here.  I haven't done a full review of this yet, but I did notice a couple things.
1. If the Cycles add-on isn't enabled, Blender will crash when baking reflection probes. (Technically it crashes on line 149 of reflection_probe.py, but I'd just disable the bake button if cycles isn't enabled and put up a warning)
2. The preferences no longer work and cause errors.  I'm not sure why you added in an enum system, but the correct way to access preferences is by `context.preferences.addons[<main package name>].preferences` and then accessing the specific preference you want through the `.` operator.  I probably should have added a function to preferences.py to easily retrieve the add-on preferences object without having to go through the long `context.preferences` etc. path and do that `__package__.split('.')[0]` thing, but it was quick and easy at the time.
3. There isn't anything to tell you what your previous reflection probes had been or to migrate the predominant size.
4. While it's very nice to have a bake all button, it would also be nice to still be able to bake probes individually, especially if you're using one reflection probe to generate an envmap and another one to apply it.

keianhzo:
> Hi, some nice improvements here. I haven't done a full review of this yet, but I did notice a couple things.
> 
>     1. If the Cycles add-on isn't enabled, Blender will crash when baking reflection probes. (Technically it crashes on line 149 of reflection_probe.py, but I'd just disable the bake button if cycles isn't enabled and put up a warning)

Good call, I tend to forget that Cycles is still just an add-on.

>     2. The preferences no longer work and cause errors.  I'm not sure why you added in an enum system, but the correct way to access preferences is by `context.preferences.addons[<main package name>].preferences` and then accessing the specific preference you want through the `.` operator.  I probably should have added a function to preferences.py to easily retrieve the add-on preferences object without having to go through the long `context.preferences` etc. path and do that `__package__.split('.')[0]` thing, but it was quick and easy at the time.

Oh yep, the preference access is wrong, fixed it. I just added a couple convenience methods to make it easier to deal with preferences.

>     3. There isn't anything to tell you what your previous reflection probes had been or to migrate the predominant size.

Yeah the previous reflection probes implementation were never officially released and this one is still a beta so I think it's ok if we don't handle migration.

>     4. While it's very nice to have a bake all button, it would also be nice to still be able to bake probes individually, especially if you're using one reflection probe to generate an envmap and another one to apply it.

Good idea, I've added a button for individual probes baking.

Exairnous:
> Good call, I tend to forget that Cycles is still just an add-on.

Yeah, me too.  What makes it worse is that if you load factory settings, it defaults to off :(

> Oh yep, the preference access is wrong, fixed it. I just added a couple convenience methods to make it easier to deal with preferences.

It's good that it's fixed, but those convenience methods are non-standard, require you to define the properties twice, and require more work to use (they require an extra import, you have two places to look through to comprehend the different preferences, and it's longer to type).  I strongly suggest getting rid of the enum and replacing `def get_addon_pref(pref):` with something like:
```
def get_addon_prefs(context):
    return context.preferences.addons[__package__].preferences
```

and using it something like this:
```
addon_prefs = get_addon_prefs(context)
style_pref = addon_prefs.my_cool_style_preference
if addon_prefs.my_bool_preference:
   set_style(style_pref)
   # etc.
```

This should result in clear, concise, standard code that is easy to use and familiar to other Blender add-on developers.

Also, best practice is to not use `bpy.context` if you can get it from a passed argument.

> Yeah the previous reflection probes implementation were never officially released and this one is still a beta so I think it's ok if we don't handle migration.

Ah, right.  I forgot that it was never officially released, and from the people I talked to at the Friday meetup, it seems like not handling migration will be fine.

> Good idea, I've added a button for individual probes baking.

Yay!  This looks generally good, but I'm curious why you removed the note about the resolution being accessible in the scene properties.  Also, this bakes all of the selected objects (which is nice), but it doesn't convey this to the user; it also fails to bake if the reflection probe is active, but isn't selected.  It might make more sense to have a bake selected button in the scene properties along with bake all, and then have the individual reflection probe just bake the active object.
Also, the warning about Cycles being disabled should be displayed on the probes as well.

keianhzo:
> It's good that it's fixed, but those convenience methods are non-standard, require you to define the properties twice, and require more work to use (they require an extra import, you have two places to look through to comprehend the different preferences, and it's longer to type). I strongly suggest getting rid of the enum and replacing `def get_addon_pref(pref):` with something like:

Fair enough. Updated.

> Yay! This looks generally good, but I'm curious why you removed the note about the resolution being accessible in the scene properties. 

I didn't, it's just displayed when a file is migrated but thinking again about it I think I can leave it permanently.

> Also, this bakes all of the selected objects (which is nice), but it doesn't convey this to the user; it also fails to bake if the reflection probe is active, but isn't selected. It might make more sense to have a bake selected button in the scene properties along with bake all, and then have the individual reflection probe just bake the active object. 

Yeah, that might be nice, I've added a "Bake selected" button.

> Also, the warning about Cycles being disabled should be displayed on the probes as well.

Fixed

Exairnous:
> > Yay! This looks generally good, but I'm curious why you removed the note about the resolution being accessible in the scene properties.
> 
> I didn't, it's just displayed when a file is migrated but thinking again about it I think I can leave it permanently.

I was referring to the fact that you changed the warning from saying "Reflection probes resolution is now set globally in scene settings." to "You can bake all reflection probes at once from the scene settings.", as an artist, one of the first things I'd look for is where my resolution settings went (or if I've never used reflection probes before, then it's a need piece of information for operation), so I think referencing where they went in the warning is needed.  I've added a potential rephrase of the warning in my latest pull request that should hopefully convey both pieces of information.

PR #89, opened 2022-07-20
Add support for shape key NLA tracks.

Exairnous:
Add support for shape key NLA tracks.
Also, fix a small api error for 2.93 and below.


keianhzo (approved):
LGTM. Great addition, thanks!

Issue #83, opened 2022-07-19
Gizmos don't update on bone duplication.


Exairnous:
**Description**
When you duplicate a bone in edit mode, no gizmo is added.

**To Reproduce**
- Add an armature.
- Add a Waypoint component to the bone.
- Enter edit mode.
- Duplicate the bone.

**Expected behavior**
The gizmo should be duplicated with the bone.

keianhzo:
Closed by https://github.com/MozillaReality/hubs-blender-exporter/pull/90

Issue #87, opened 2022-07-19
Gizmos don't undo/redo properly on gizmo component removal


Exairnous:
**Description**
If you undo removing a gizmo component, the gizmo doesn't reappear.
If you undo adding a gizmo component and then redo, the gizmo doesn't reappear

**To Reproduce**
- Add an object.
- Add a Waypoint component to the object.
- Remove the Waypoint component from the object.
- Undo.
- Delete the object.
- Add another object.
- Add a Waypoint component to the object.
- Undo.
- Redo.

**Expected behavior**
The gizmo should be present on the object if the component is.

keianhzo:
Closed by https://github.com/MozillaReality/hubs-blender-exporter/pull/90

Issue #86, opened 2022-07-19
Gizmo disappears on host rename


Exairnous:
**Description**
Renaming the object/bone the gizmo is attached to causes the gizmo to disappear.

**To Reproduce**
- Add an object.
- Add a Waypoint component to the object.
- Rename the object.

**Expected behavior**
The gizmo should remain visible on the object.

keianhzo:
Closed by https://github.com/MozillaReality/hubs-blender-exporter/pull/90

Issue #85, opened 2022-07-19
Gizmos on bones on other armatures don't show up correctly when in edit mode


Exairnous:
**Description**
If you enter edit mode on one armature all the gizmos on other armatures will disappear.

**To Reproduce**
- Add an armature.
- Duplicate the armature and move it to a new location.
- Rename the bone in the new armature to a unique name.
- Add a Waypoint component to both bones.
- Enter edit mode on one of the armatures.

**Expected behavior**
The gizmos should show on both bones.

keianhzo:
Closed by https://github.com/MozillaReality/hubs-blender-exporter/pull/90

PR #90, opened 2022-07-21
Fix Gizmo Issues

Exairnous:
This should fix all the issues with gizmos, plus improve the performance and prevent any performance drain on scenes that aren't using gizmos.  I feel it also simplifies the logic in some places.  Objects are now linked directly with gizmos for fast lookups and shouldn't become stale because the system is re-initialized on any change, this also means that there is no further use of a dedicated remove gizmo function, so it has been removed.  Blender's msgbus system has been employed to trigger a refresh when a gizmo object is renamed (unfortunately bones seem to be too unstable to be subscribed to directly, so a refresh is triggered on any bone rename)

This PR fixes issues:
#83
#84
#85
#86
#87

### DIFF
--- a/addons/io_hubs_addon/components/gizmos.py
+++ b/addons/io_hubs_addon/components/gizmos.py
@@ -1,5 +1,4 @@
 import bpy
-import bpy_types
 from bpy.types import (Gizmo, GizmoGroup)
 from bpy.props import (IntProperty)
 from .components_registry import get_component_by_name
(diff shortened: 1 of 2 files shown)


keianhzo (approved):
LGTM. I haven't experimented with the msg bus too much but it seems reasonable way of conditionally updating objects. Good improvement, thanks!

keianhzo:
BTW I just realized that the related issues haven't closed after merging this. I will close them manually. For future reference if you want to close them as a result of merging the PR you have to put `Close` in front: 

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #94, opened 2022-07-26
Additional reflection probe fixes and improvements.

Exairnous:
I found a couple issues when finishing my review of the new reflection probe functionality.  The biggest issues I found were that baking always used the default and that the render resolution percentage wasn't accounted for i.e. set to 100% before baking and restored afterward.  So I decided to fix the issues and also took the opportunity to simplify some logic and improve the performance in a few spots.  One of the places I reworked was the probe selection based on the mode, logical simplifications like assuming that all reflection probes are also light probes (since that's the only object they can be added to) and that the active object will be correct if you can see the bake button on the active object, stuff like that (I assume this should be fine since we aren't handling migration).  See below for a full list of changes.

Remove double import.
Keep a reference to the strings returned for the env_map EnumProperty item.
Add a default value to reference_id.
Simplify the EnumProperty handling.
Use the chosen resolution when baking, instead of always using the default.
Simplify the mode lookup for the bake operator.
Account for the render resolution percentage.
Include the fact that the resolution has moved in the warning.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/reflection_probe.py
+++ b/addons/io_hubs_addon/components/definitions/reflection_probe.py
@@ -1,5 +1,5 @@
 import bpy
-from bpy.props import PointerProperty, EnumProperty, StringProperty, StringProperty
+from bpy.props import PointerProperty, EnumProperty, StringProperty
 from bpy.types import Image, PropertyGroup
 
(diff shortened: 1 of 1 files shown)


keianhzo (approved):
LGTM

PR #97, opened 2022-07-29
Fix Loop Animation Component Issues

### LINKED ISSUE #93: Loop Animation Component Issues
**Description**
List of UI/interaction issues with the Loop Animation component in the 1.0.0 beta:
- Cannot add tracks to the object component when in pose mode.
- Cannot add tracks to the bone component when in object mode.
- Cannot remove tracks from the bone component when in object mode.
- Cannot remove tracks from the bone component if a track from the object mode component isn't selected.
- Cannot remove tracks from the object component when in pose mode.
- Can add duplicate tracks to the bone component when in pose mode from either panel.
- The remove operator isn't disabled when the last track is removed.
- The remove operator isn't disabled on bones if there is a track selected on the object.
- The added track isn't selected on creation.
- The track selection becomes invalid if you select the last track and then delete it.
- Cannot undo/redo adding or removing a track.

**To Reproduce**
- Add a Loop Animation component to an object or bone (or both).
- Attempt to do one of the things listed above.

**Expected behavior**
The Loop Animation component should take the panel context into account and behave like similar UI elements in Blender.

**Hardware** 
 - Add-on Release Version: 1.0.0 beta
 - Blender Version: 2.93.4, 3.1.2, 3.2.1

Exairnous:
Fixes #93 

Uses the panel type to specify which component (object or bone) the operators should act on.
`context_pointer_set` is used to pass the panel to the menu and the remove track operator (the poll function can't access operator arguments).

The item selection behavior has been changed to match the default behavior.

I've also changed the new draw_global method to take the panel instead of the panel type in case it needs to be passed via  `context_pointer_set` in the future.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/loop_animation.py
+++ b/addons/io_hubs_addon/components/definitions/loop_animation.py
@@ -50,7 +50,8 @@ class AddTrackOperator(Operator):
         track = host.hubs_component_loop_animation.tracks_list.add()
         track.name = self.track_name
 
-        host.hubs_component_loop_animation.active_track_key += 1
+        num_tracks = len(host.hubs_component_loop_animation.tracks_list)
+        host.hubs_component_loop_animation.active_track_key = num_tracks - 1
 
         return {'FINISHED'}
 
--- a/addons/io_hubs_addon/components/definitions/loop_animation.py
+++ b/addons/io_hubs_addon/components/definitions/loop_animation.py
@@ -35,17 +35,22 @@ class TracksList(bpy.types.UIList):
 class AddTrackOperator(Operator):
     bl_idname = "hubs_loop_animation.add_track"
     bl_label = "Add Track"
+    bl_options = {'REGISTER', 'UNDO'}
 
     track_name: StringProperty(
         name="Track Name", description="Track Name", default="")
 
-    def execute(self, context):
-        ob = context.object
-        if context.mode == 'POSE' or context.mode == 'EDIT_ARMATURE':
-            ob = context.active_bone
+    panel_type: StringProperty(name="panel_type")
 
-        track = ob.hubs_component_loop_animation.tracks_list.add()
+    def execute(self, context):
+        panel_type = PanelType(self.panel_type)
+        ob = context.object
+        host = ob if panel_type == PanelType.OBJECT else context.active_bone
+
+        track = host.hubs_component_loop_animation.tracks_list.add()
         track.name = self.track_name
+
+        host.hubs_component_loop_animation.active_track_key += 1
 
         return {'FINISHED'}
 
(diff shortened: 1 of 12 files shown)


addons/io_hubs_addon/components/definitions/loop_animation.py
```diff
@@ -35,18 +35,24 @@ def draw_item(self, context, layout, data, item, icon, active_data, active_propn
...
     bl_label = "Add Track"
+    bl_options = {'REGISTER', 'UNDO'}
 
     track_name: StringProperty(
         name="Track Name", description="Track Name", default="")
 
+    panel_type: StringProperty(name="panel_type")
+
     def execute(self, context):
+        panel_type = PanelType(self.panel_type)
```
keianhzo, line 46: Once the `context_pointer_set` has been set for the layout can't we use the panel here same as we are doing for `RemoveTrackOperator` and get rid of the property?

addons/io_hubs_addon/components/definitions/loop_animation.py
```diff
@@ -88,20 +105,26 @@ class TracksContextMenu(Menu):
...
         if ob.animation_data:
             for _, a in enumerate(ob.animation_data.nla_tracks):
-                if not has_track(context.object.hubs_component_loop_animation.tracks_list, a.name):
-                    self.layout.operator(AddTrackOperator.bl_idname, icon='OBJECT_DATA',
-                                         text=a.name).track_name = a.name
+                if not has_track(host.hubs_component_loop_animation.tracks_list, a.name):
+                    add_track = self.layout.operator(AddTrackOperator.bl_idname, icon='OBJECT_DATA',
+                                         text=a.name)
+                    add_track.track_name = a.name
+                    add_track.panel_type = panel_type.value
```
keianhzo, line 118: If we use the context panel pointer we can get rid of this.

addons/io_hubs_addon/components/definitions/loop_animation.py
```diff
@@ -88,20 +105,26 @@ class TracksContextMenu(Menu):
...
 
         if hasattr(ob.data, 'shape_keys') and ob.data.shape_keys and ob.data.shape_keys.animation_data:
             for _, a in enumerate(ob.data.shape_keys.animation_data.nla_tracks):
                 if not has_track(context.object.hubs_component_loop_animation.tracks_list, a.name):
-                    self.layout.operator(AddTrackOperator.bl_idname, icon='OBJECT_DATA',
-                                         text=a.name).track_name = a.name
+                    add_track = self.layout.operator(AddTrackOperator.bl_idname, icon='OBJECT_DATA',
+                                         text=a.name)
+                    add_track.track_name = a.name
+                    add_track.panel_type = panel_type.value
```
keianhzo, line 127: Same here

keianhzo (changes requested):
Looks good. Just a couple of nits.

addons/io_hubs_addon/components/definitions/loop_animation.py
Exairnous: Yes, that's possible, and I did consider it, but I think having the arguments is clearer than using `context_pointer_set`.  The only reason I used it for the `RemoveTrackOperator` was because I needed to access it in the `poll` function and you can't access the arguments from there.  That said, we are using it for the `RemoveTrackOperator`, so I don't have a really strong feeling about this and I'm fine with changing it to use `context_pointer_set` if you want.

keianhzo: Ok, I see. I was just curious why but I'm ok with leaving it like it is.

keianhzo (approved):

PR #98, opened 2022-07-29
Fix morph-audio-feedback initial missing shape key warning

Exairnous:
Fix the erroneous warning being shown initially in the morph-audio-feedback component's Shape Key menu.


keianhzo (approved):
LGTM

PR #99, opened 2022-08-01
Support default NLA Track names

Exairnous:
**This depends/is built on the Fix Loop Animation Component Issues PR**

Add support for default NLA Track names and improve the component's UI.

Handle the glTF exporter replacing the default track names (NlaTrack and [Action Stash]) with the action names, and make sure this is nicely conveyed by the UI.

Improve the component UI by redrawing immediately on changes, using the shape key icon for shape key tracks, and making the user select a new track from a menu instead of typing the new name in themselves (this is required to handle default track names).

Improve error reporting slightly, now tracks that contain multiple actions, no actions, actions that have since been renamed, or have names (track or action depending on whether the track name is a default) containing forbidden characters, will display with error icons.

Note: Invalid tracks can still be added as long as they aren't already present. Duplicate tracks will only be shown once in the list (as they will be combined on export).  Duplicate tracks are determined based on track name (if it's not a default name) or action name (if it is).  If you have duplicate object and shape key tracks, only the object track will be shown in the add menu.

Note2: Any files saved on a previous beta are not migrated automatically; users will need to update or re-add their tracks (the same as if the track names had changed).


addons/io_hubs_addon/components/definitions/loop_animation.py
```diff
@@ -1,9 +1,160 @@
...
+    nla_track_name_msgbus_owner = None
+    action_name_msgbus_owner = None
+
+@persistent
+def load_post(dummy):
+    unregister_msgbus()
+    register_msgbus()
+
+def is_default_name(track_name):
+    return bool(track_name.startswith("NlaTrack") or track_name.startswith("[Action Stash]"))
```
keianhzo, line 84: I'm glad that we have tests as this is a good candidate for future breakage.

addons/io_hubs_addon/components/definitions/loop_animation.py
```diff
@@ -1,9 +1,160 @@
...
+            ("shape_key", "Shape Key", "Shape Key")
+            ],
+        default="object"
+    )
+
+bpy.utils.register_class(TrackPropertyType)
+
+@atexit.register
+def unregister():
+    bpy.utils.unregister_class(TrackPropertyType)
```
keianhzo, line 40: I've landed a PR that changes how we handle module-wide class registration. It's a pretty simple fix, you just need to move this to a module `register_module`/`unregister_module` functions.

keianhzo (changes requested):
This is looking great and it's making user's life much easier! Just a thing that needs to be updated based on a recent update that I did.

Exairnous:
> This is looking great and it's making user's life much easier! Just a thing that needs to be updated based on a recent update that I made.

Yup, I think just merging in master should solve that.  Using atexit was a bit weird, so I'm glad you've replaced it, but it's still a little odd having two register functions, so I may look into it later and see if we can get that down to just one.

> Regarding the test error, as we talk in the chat probably the best solution is just to update the original test blend file to properly test both the named and default-named NlaTrack cases.

Okay, yep.  Will do.

addons/io_hubs_addon/components/definitions/loop_animation.py
Exairnous: Yes, definitely.

Exairnous:
Okay, I've merged in master and updated the tests.  I didn't do anything about checking whether the present animation matched what's in clip though (for the tests).  I think everything should be good now, but I'm not very familiar with unit tests, so let me know if any changes are needed.

Exairnous:
I found a couple serious issues with this, so I'm marking it as a draft until I get them fixed.
I got confused and thought strips were actions, so I was using the strip name, rather than the strip's action's name, plus I found out that the way the glTF exporter handles exporting animations is even weirder than I thought, so I'm not handling duplicate actions correctly.

keianhzo:
Closing in favor of https://github.com/MozillaReality/hubs-blender-exporter/pull/100

PR #100, opened 2022-08-01
Add error messages to the update track menu.

Exairnous:
**This depends/is built on the Support default NLA Track names PR**

This is a test for displaying granular error messages.  Since we haven't talked much about this I've added it as a separate PR so we can experiment if we like, but I quite like what I have here.

Use the update track menu to display granular error messages.

This includes error messages for:
Tracks that aren't found (i.e. tracks that have been renamed or deleted).
Names containing forbidden characters.
Tracks that contain multiple actions.
Tracks that don't contain any action.
Tracks that contain an action that no longer has a corresponding action in Blender (i.e. the action has been renamed or deleted, but the track is still referencing the old action).

A list of other tracks is only displayed in the menu if it's a track not found error, otherwise it just shows the error.


keianhzo (commented):
I think it's nice to show information about errors and I like the menu to quickly show alternatives when an error is found so the user can quickly fix it. 

For the error messages I still find more usable the red labels that we are using in other components. It's easier to spot and read and also that would keep a consistency in the way we flag errors in the add-on.

My preferred path forward for this would be a combination of a red label warning right under the tracks panel and the error icon + track menu offering possible track alternatives.

Exairnous:
Hm, are you suggesting adding a generic red error underneath the list that says something like "Errors detected, click on the flagged tracks for more information."?  I originally thought of putting the errors underneath in red, but then discarded the idea because multiple errors can be present at the same time, so I ended up needing to specify which track had the error and then that didn't leave much room for the actual error message.  However, having an additional generic error underneath might work really well.  It would provide notification if the invalid track is scrolled out of view and instruction on where to go for more detailed information.  The more I think about it, the more I like it.

I've only included alternative tracks in the menu if it's a track not found error, because all the other errors are NLA track or action issues that you need to fix elsewhere.  I feel that having the alternative tracks present in these cases is just distracting and not really useful.  That said, this is all experimental, so if you want to try having the tracks present all the time, or anything else, I'm happy to oblige.

keianhzo:
> Hm, are you suggesting adding a generic red error underneath the list that says something like "Errors detected, click on the flagged tracks for more information."? 

Yeah, I just think that the user might benefit from UI consistency. We don't need to add all the details, what you proposed above sounds good to me.

Exairnous:
I've added in the generic message to test.  I haven't merged in any of the other changes from master or the base 'support generic names' branch because there haven't been any relevant functional changes and I think it's probably better to do that at the end before merging.

Exairnous:
I fixed everything with the strip/action names and improved the error reporting, so now everything feels pretty solid.  There's a bit more micromanaging involved than I would like, but I couldn't find a good way around that and I think this is the best way to provide a stable, predictable workflow for the user.  I had initially separated out the default name support and the detailed error reporting, but given the increase in complexity, I now don't think the default name support should be merged without the error messages.  If you're agreeable, I'd like to just move forward with this one and you can close #99.

keianhzo (approved):
LGTM

PR #106, opened 2022-08-24
Disable compostior and don't clobber settings during reflection probe baking

netpro2k:
This implements 2 todos that were sitting in the reflection probes code that are leading to much pain for artists:
- Disable compositor during reflection probe baking
- Don't clobber render settings

With this fix it should not matter what the user has their render settings set to before baking. Also, after baking their render settings should be restored to what they were before the bake.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/reflection_probe.py
+++ b/addons/io_hubs_addon/components/definitions/reflection_probe.py
@@ -10,6 +10,7 @@ from ..components_registry import get_components_registry
 from ..hubs_component import HubsComponent
 from ..types import Category, PanelType, NodeType
 from ... import io
+from ...utils import rgetattr, rsetattr
 import math
 
 
@@ -268,10 +269,16 @@ class BakeProbeOperator(bpy.types.Operator):
         self.camera_object.rotation_euler.x += math.pi/2
         self.camera_object.rotation_euler.z += -math.pi/2
 
-        bpy.context.scene.camera = self.camera_object
-        bpy.context.scene.render.engine = "CYCLES"
         resolution = context.scene.hubs_scene_reflection_probe_properties.resolution
         (x, y) = [int(i) for i in resolution.split('x')]
+
+        saved_props = {}
+        props_to_save = ["camera", "render.engine", "cycles.device", "render.resolution_x", "render.resolution_y", "render.resolution_percentage", "render.image_settings.file_format", "render.filepath", "render.use_compositing"]
+        for prop in props_to_save:
+           saved_props[prop] = rgetattr(bpy.context.scene, prop)
+
+        bpy.context.scene.camera = self.camera_object
+        bpy.context.scene.render.engine = "CYCLES"
         bpy.context.scene.cycles.device = "GPU" if is_gpu_available(
             context) else "CPU"
         bpy.context.scene.render.resolution_x = x
@@ -280,9 +287,7 @@ class BakeProbeOperator(bpy.types.Operator):
         bpy.context.scene.render.image_settings.file_format = "HDR"
         bpy.context.scene.render.filepath = "%s/%s.hdr" % (
             get_addon_pref(context).tmp_path, probe.name)
-
-        # TODO don't clobber renderer properties
-        # TODO handle skipping compositor
+        bpy.context.scene.render.use_compositing = False
 
         tmp_pref = bpy.context.preferences.view.render_display_type
         bpy.context.preferences.view.render_display_type = "NONE"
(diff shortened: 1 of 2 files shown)


Exairnous:
Hey Dom, great minds think alike.  I was also thinking about adding in support for bypassing the compositor :)

There's actually already support for not clobbering the render settings (but it seems no one removed the todo), although this might be a bit nicer.  Unfortunately, what you have doesn't render via the correct camera though.  The problem is that `bpy.ops.render.render("INVOKE_DEFAULT", write_still=True)` is a modal operator (it's asynchronous), so with the way you're doing it now, you end up resetting all of the overrides immediately after the render starts (this probably also caused the problem with the compositor toggle you mentioned in chat).  The good news is that I've found a pretty easy fix:

1. Make `saved_props` a class attribute of the `BakeProbeOperator` Operator.
2. Only save the initial settings:
```python
for (prop, value) in overrides:
    if prop not in self.saved_props:
        self.saved_props[prop] = rgetattr(bpy.context, prop)
```
3. Replace the contents of `restore_render_props` method with this and remove it from the end of the `render_probe` method:
```python
for prop, value in self.saved_props.items():
    rsetattr(bpy.context, prop, value)
```
4. Remove this from the `execute` method, as it is no longer necessary with your new code:
```python
self.prev_render_camera = bpy.context.scene.camera
self.prev_render_engine = bpy.context.scene.render.engine
self.prev_cycles_device = bpy.context.scene.cycles.device
self.prev_render_rex_x = bpy.context.scene.render.resolution_x
self.prev_render_res_y = bpy.context.scene.render.resolution_y
self.prev_render_res_percent = bpy.context.scene.render.resolution_percentage
self.prev_render_file_format = bpy.context.scene.render.image_settings.file_format
self.prev_render_file_path = bpy.context.scene.render.filepath
```

You could probably also remove the latter two `self.restore_render_props()` calls in the `modal` method, as I think only the first one is necessary, but I haven't tested that to make sure.

And a final nit, it looks like you've indented your blocks with three spaces instead of the standard four.

keianhzo:
Oh yep, I forgot to remove the clobber comment. So basically what @Exairnous said, we just need to set the compositor to false and save/restore that previous value.

My only request would be to add a checkbox to enable/disable this as some users might want to use the compositor even for baked reflection probes. I can be disabled by default to avoid issues.

Exairnous:
I think this works well now, but I do agree with @keianhzo that there should be an option to allow the compositor to be enabled for reflection probes.  Aside from that, I just noticed "Render Region" and "Crop to Render Region" could use overriding as well.

And a couple nits/questions/suggestions,
1. Is there a reason you and @keianhzo don't like using `items()` when looping through dictionaries?
2. This doesn't affect the functionality this time, but I noticed that you made saved_props an instance attribute, rather than a class attribute, and this got me to thinking that we should probably standardize on one for this type of thing, and that what we probably want are instance attributes.

netpro2k:
> And a couple nits/questions/suggestions,
> 
>   1.  Is there a reason you and @keianhzo don't like using items() when looping through dictionaries?
>    2. This doesn't affect the functionality this time, but I noticed that you made saved_props an instance attribute, rather than a class attribute, and this got me to thinking that we should probably standardize on one for this type of thing, and that what we probably want are instance attributes.

Both of these (for me) come down to the fact that I only know enough Python to be dangerous :). Definitely open to any suggestions on idiomatic Python code when we are doing something odd.

Exairnous:
>> there should be an option to allow the compositor to be enabled for reflection probes
> 
> That seems fine. It seems to more common case is that having the compositor apply is surprising, so that's what I am trying to fix in this PR.

I think the reported cases have been in line with that, but it would be really nice to be able to denoise them or apply a certain look/filter/etc. automatically (Actually, it would be nice to be able to have a specific compositor node tree for reflection probes and or individual probes, but that, while it should be possible, would be a bit more involved ;) ).

> Both of these (for me) come down to the fact that I only know enough Python to be dangerous :). Definitely open to any suggestions on idiomatic Python code when we are doing something odd.

Ah, okay :).
I don't think there's any hard and fast rule for the `items()` thing, although it may be slightly more Pythonic to use `items()`, but using `items()` is more explicit and by retrieving both the key and the value in the for loop you don't have to lookup the value later (you also don't have to remember that looping through the dictionary without a "suffix" loops through the keys rather than the values, which is something I always forget).

For class attributes vs instance attributes, the differences are subtle and can cause obscure bugs if you're not careful.  Instance attributes are declared via `self.attribute_name = value` in a class method and are local to that instance (often times they're done in `__init__`, but Blender doesn't really do `__init__`, so `execute` or `invoke` are other good places), while class attributes are declared beneath the class heading (outside of any method) and are global to the class, provided they are accessed/set via `classname.attribute_name` if they are accessed/set via `self.attribute_name` from within a method their values will be local to that instance, unless they have a mutable value (dictionary, list, etc.) that hasn't been reassigned.  (I encourage looking this up in more detail as I'm probably forgetting some intricacies)

keianhzo (approved):
LGTM

PR #107, opened 2022-08-25
Fixes references on node properties

### LINKED ISSUE #81: Blender 3.2: Video Texture Target doesn't work in Hubs
**Description**
The video texture target component doesn't work in Hubs when exported from Blender 3.2.

**To Reproduce**
- Start with Blender's default scene.
- Add a Video Source Component to the camera.
- Add a Video Texture Target component to the default cube's material.
- Add a new image texture to the material's base color.
- Set the Video Texture Target's source property to the camera.
- Export to glTF.
- In the exporter window, under the include panel, check the checkbox for Cameras.
- Press 'Export glTF 2.0'.
- Drop the glb into a Hubs room.

**Expected behavior**
That what the camera sees will be rendered to the cube.

**Hardware** 
 - Add-on Release Version: 1.0.0 Beta
 - Blender Version: 3.2.0, 3.2.1

**Additional context**
The console in Chromium provides these warnings:
- inflateComponents: invalid node reference srcNode
- Error inflating gltf component "video-texture-srcEl": Couldn't find srcEl entity with index null

From examination of the glTF file, it seems like the camera is getting included twice in the "nodes" list.
[untitled-3.1.gltf.txt](https://github.com/MozillaReality/hubs-blender-exporter/files/9068876/untitled-3.1.gltf.txt)
[untitled-3.2.gltf.txt](https://github.com/MozillaReality/hubs-blender-exporter/files/9068877/untitled-3.2.gltf.txt)

The same issue is present if a bone is used as the source.

sarhugo:
The vTree is pregenerated so it should already have all the node references. For each reference it was creating a new node causing duplicates at export. Previous version of gltf-exporter (3.1) had a cache that prevented this.

This should fix #81

### DIFF
--- a/addons/io_hubs_addon/components/definitions/audio_target.py
+++ b/addons/io_hubs_addon/components/definitions/audio_target.py
@@ -4,7 +4,7 @@ from ..hubs_component import HubsComponent
 from ..utils import has_component
 from ..types import Category, PanelType, NodeType
 from bpy.types import Object
-from ...io.utils import gather_joint_property, gather_node_property
+from ...io.utils import gather_joint_property, gather_node_property, delayed_gather
 
 BLANK_ID = "374e54CMHFCipSk"
 
@@ -151,6 +151,7 @@ class AudioTarget(HubsComponent):
         layout.prop(data=self, property="maxDelay")
         layout.prop(data=self, property="debug")
 
+    @delayed_gather
     def gather(self, export_settings, object):
         return {
             'srcNode': gather_joint_property(export_settings, self.srcNode, self, 'bone') if self.bone_id != BLANK_ID else gather_node_property(
--- a/addons/io_hubs_addon/components/definitions/video_texture_target.py
+++ b/addons/io_hubs_addon/components/definitions/video_texture_target.py
@@ -3,7 +3,7 @@ from ..hubs_component import HubsComponent
 from ..types import Category, PanelType, NodeType
 from ..utils import has_component
 from bpy.types import Object
-from ...io.utils import gather_joint_property, gather_node_property
+from ...io.utils import gather_joint_property, gather_node_property, delayed_gather
 
 BLANK_ID = "pXph8WBzMu9fung"
 
@@ -141,6 +141,7 @@ class VideoTextureTarget(HubsComponent):
             col.label(text='This component requires a material',
                       icon='ERROR')
 
+    @delayed_gather
     def gather(self, export_settings, object):
 
         return {
--- a/addons/io_hubs_addon/io/gltf_exporter.py
+++ b/addons/io_hubs_addon/io/gltf_exporter.py
@@ -71,6 +71,7 @@ class glTF2ExportUserExtension:
         self.Extension = Extension
         self.properties = bpy.context.scene.HubsComponentsExtensionProperties
         self.was_used = False
+        self.hubs_node_property_references = {}
 
     def hubs_gather_gltf_hook(self, gltf2_object, export_settings):
         if not self.properties.enabled or not self.was_used:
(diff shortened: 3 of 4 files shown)


keianhzo:
Good catch, thanks for the pointer. Unfortunately I think this patch doesn't really work in all cases. As you mentioned the vtree is populated but it is as nodes are traversed so depending on the order the node the material is pointing at might not yet have been traverse and thus is not linked in the vtree.

The only workaround for > 3.2.0 that I can think of is to traverse materials in `gather_scene_hook` after all nodes are already available and create the `srcNodes`.

sarhugo:
Yes, I didn't think about the processing order, you are right. Ok, I'll check how to adapt it, in summary:

- put the referenced object instead of a node at gather_node_property.
- in gather_scene_hook I have to traverse every node, bone, material and the scene itself to check for any component attached that has a node reference and then correctly attach the node from already populated vtree

sarhugo:
Or ask the guys from gltf exporter to reenable the cache :shrug:

Exairnous:
Hm, very interesting.  I didn't know this happened.
I think processing the nodes in the scene hook, after they've all been generated, is a good idea, but I think it might be nicer if you could store references to the delayed nodes (or the gather_imports that gets nodes) and then iterate over just those.  I have a couple vague ideas about how you might go about doing this, possibly involving decorators or storing functions, but I'll have to get back to you on this.

I don't think asking for the cache to be re-enabled on the glTF exporter is really viable at this point because it won't do anything for Blender 3.2 or 3.3 (wouldn't make it in time for release).

sarhugo:
Made the logic to process every node property reference at scene_hook... not sure if the place where I stored the references is ok, maybe you have a better way of doing it

Exairnous:
@sarhugo I think where you have stored the references is okay.  While this seems to work nicely and may end up being the best solution for this problem, it also seems fairly complicated and somewhat hard to reason about (e.g., I'm uncertain how `self.hubs_node_property_references` is getting filled?).  I've thought about this problem some more and an alternate method, which is also more flexible for delaying component gathering should we need it for something else, is to use a decorator to wrap the gather functions like so:

```python
def delayed_gather(func):
    def gather_wrapper(self, export_settings, object):
        def gather():
            return func(self, export_settings, object)
        gather.delayed_gather = True
        return gather
    return gather_wrapper
```

This will cause a gather function decorated with this to store the arguments and return a function with a `delayed_gather` attribute, then in `add_hubs_components` you can simply do this:

```python
data = component.gather(
    export_settings, blender_object)
if hasattr(data, "delayed_gather"):
    self.delayed_gathers.append((component_data, component_class.get_name(), data))
else:
    component_data[component_class.get_name()] = data
```

And then loop through `self.delayed_gathers` in the `gather_scene_hook` and gather the data that needed the delay and add it to the component data for export:

```python
for delayed_gather in self.delayed_gathers:
    component_data, component_name, gather = delayed_gather
    component_data[component_name] = gather()
self.delayed_gathers.clear()
```

This all works nicely because Python passes dictionaries by reference, so you can store and modify them and anything that references them will get your updates, i.e. you don't have to worry that you skipped adding the component data for the delayed gathers in `add_hubs_components` because you add them to the stored reference (which is always up to date and will propagate your changes to anything else that uses it) in the `gather_scene_hook`. 

With this all in place, any `gather` function decorated with `delayed_gather` will end up running at the end of `gather_scene_hook` and so all you need to do then is your initial modification of `vnode.node or gltf2_blender_gather_nodes.gather_node` to `gather_node_property` and `gather_joint_property`.

Note: if we wanted to make the decorator even more generic (e.g. for delaying import gathers as well), I think we could replace the arguments `(self, export_settings, object)` with `(*args, **kwargs)` and that should allow it to delay any function in this way (I haven't tested this, but I think it should work).

sarhugo:
@Exairnous check the latest commit... thanks for all the perfect step by step explanation, makes totally sense

Exairnous:
You're welcome!  @keianhzo will need to sign off on it, but this looks good to me :)

keianhzo (approved):
This looks great! I really like the decorator approach, makes things much simpler.

PR #108, opened 2022-09-01
Don't modify viewpoert visibility from Visible component

netpro2k:
The behavior seems to be more confusing then helpful, so just removing it for now. We can explore other useful ways of indicating that this component is on an object if we find it necessary.


keianhzo (approved):
Yeah, this can make things more confusing. LGTM

PR #109, opened 2022-09-01
Add startOffset and timeScale to loop-animation

netpro2k:
Adds `startOffset` (skip this many seconds on the first loop of the animation) and `timeScale` (scale the animation playback speed by this factor) to the "loop-animation" component. Note in the blender UI `startOffset` will be displayed in frames, converted to seconds at export based on the scene's FPS settings.

Also removed "paused" as it doesn't actually work, and doesn't make much sense given the lack of any way to actually start the animation programmatically if it did work... 

Requires Hub change https://github.com/mozilla/hubs/pull/5692

### DIFF
--- a/addons/io_hubs_addon/components/definitions/loop_animation.py
+++ b/addons/io_hubs_addon/components/definitions/loop_animation.py
@@ -1,6 +1,6 @@
 import bpy
 from bpy.app.handlers import persistent
-from bpy.props import StringProperty, CollectionProperty, IntProperty, BoolProperty, EnumProperty
+from bpy.props import StringProperty, CollectionProperty, IntProperty, BoolProperty, EnumProperty, FloatProperty
 from bpy.types import PropertyGroup, Menu, Operator
 from ..hubs_component import HubsComponent
 from ..types import Category, PanelType, NodeType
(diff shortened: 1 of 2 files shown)


Exairnous:
Does "paused" not work with Trigger Volume?  I am hoping to add Trigger Volume to the exporter in the nearish future.

keianhzo (approved):
LGTM

netpro2k:
Woops should have marked this as a draft. Still waiting for some test scenes to actually try it in Hubs. I suspect it will be fine though.

Exairnous:
> > Does "paused" not work with Trigger Volume? I am hoping to add Trigger Volume to the exporter in the nearish future.
> 
> Not sure if we want to add volume trigger support in the Blender Add-on so I think removing paused is ok for now.

I have had several requests from artists for it to be supported and I was planning to open a PR for it soon after the add-on was out of beta.

netpro2k:
We certainly want some form of trigger volume thing, but the current thing that exists was a total experiment and is not really supported in its current form ( I suspect its also broken in all sorts of ways). I don't think we should bother adding it to the addon in its current form since that just adds more work whenever we do come up with the "real" thing.

Exairnous:
Yeah, I understand it's not really supported and will be eventually replaced with a better version, but as long as it works to some degree and is useful to artists, and until it's actually removed I think it should be added to the add-on, otherwise people are still going to have juggle scenes between the add-on and Spoke.  With the current architecture, it's really not difficult to support and when the time comes it can just be deleted and replaced with the "real" thing.  I'm happy to make this my pet project and assume responsibility for it.

However, if everyone is really against having it in the main add-on, perhaps a repo of experimental definitions could be created by the Hubs team and it could live there and people could then download it and add it to their local copy of the add-on if they wish?  As a sort of middle ground?

netpro2k:
One of the things we wanted with the new addon architecture was the ability for people to be able to specify a folder/folders for their own components. This is how we would want third party devs to work with their own components.

Exairnous:
So, are you saying that you would prefer my "middle ground" proposal, but want to wait until the external definition loading is present?

Exairnous:
Another thing that just occurred to me about having the paused property on loop animation is that I've been thinking for awhile now that it would be really cool to be able to toggle an avatar's animations on/off with a keyboard shortcut, so if that ever gets implemented then we'd want the paused property.  Obviously this doesn't affect anything currently, but it was on my mind, so I thought I'd mention it.

Issue #92, opened 2022-07-22
Show/Hide audio-params component audio properties depending on the selected audio type/distance model


keianhzo:
**Is your feature request related to a problem? Please describe.**
Not all the `audio-params` component properties are used for every audio type and distance model. In Spoke we are already hiding some properties depending on those. We should do the same in the Blender add-on.

**Describe the solution you'd like**
Update the displayed properties based on the selected audio type and distance model.

**Additional context**
Take a look at Spoke for reference about what properties should be displayed depending on the audio type and distance model.

PR #110, opened 2022-09-09
Conditional UI for audio-params properties

### LINKED ISSUE #91: Support audio parameters override
**Is your feature request related to a problem? Please describe.**
In Spoke we support conditionally overriding the audio parameters of those components that had a `audio-params` dependency. If the parameters are not overridden it will use the platform defaults. 

**Describe the solution you'd like**
We should have a checkbox in the `audio-params` component to enable/disable it. If disabled the component won't be written to the GLTF file.

sarhugo:
Only show properties and export component if override settings is selected.

Also show available properties according to audio type.

Fixes #91 #92


keianhzo (approved):
LGTM. Sorry that this took so long to review and thanks for the contribution!

PR #114, opened 2022-09-11
Set the cutoff version for using "children_recursive" to Blender 3.1 

Exairnous:
Blender 3.0 doesn't have the "children_recursive" convenience function.


keianhzo (approved):
LGTM

PR #115, opened 2022-09-11
Add option to reflection probes to use the compositor.

Exairnous:
Add the option for reflection probes to send the baked images through the compositor after baking.


keianhzo (approved):
LGTM

PR #116, opened 2022-09-11
Remove unneeded "soft_min" and "soft_max" parameters.

Exairnous:
Having the same "soft" min/max as "hard" min/max is redundant. "soft_min" and "soft_max" are used to set guidelines for the user, i.e. if used on a slider they will be unable to go past these values when clicking and dragging, but this doesn't limit what they can type in directly (that's what min and max do, provide a hard limit that you can't bypass).


keianhzo (approved):
LGTM

PR #117, opened 2022-09-11
Fix typos, improve docs, and remove trailing periods.

Exairnous:
Fixed typos and grammar errors.
Added the full (mozilla.org) default URL to the video component (it was just https://). Improved the descriptions/tooltips to better document what the things do. Removed the trailing periods from descriptions/tooltips as Blender provides them for us and will cause duplicates if we provide them too. Periods aren't provided for labels, so those have been left.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ You can override the Blender user scripts directory from the console to point to
 
 `BLENDER_USER_SCRIPTS=full_path_to_/hubs-blender-exporter blender`
 
-- ### Symlinking you addon to the Blender user scripts directory
+- ### Symlinking your addon to the Blender user scripts directory
 You can create a symbolic link pointing to `full_path_to_/hubs-blender-exporter/addons/io_hubs_addon` in your current Python scripts directory in Blender. This way you will also load any other addons that you have in that directory.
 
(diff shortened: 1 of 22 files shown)


sarhugo:
Maybe here is a place to fix this? https://github.com/MozillaReality/hubs-blender-exporter/issues/77

Exairnous:
@sarhugo Yup, Thank you!  That should definitely be in here!

Exairnous:
Actually, I may have spoken too soon, this could be a little tricky because not just the display name and tooltip is misspelled.  I think the Hubs client would have to be modified to accept the new spelling for the component name as well, plus we'll need to provide a migration for old blend files.  All that to say that it's probably too big for this PR, unless we want to keep the component name and just fix the display name and tooltip.  @keianhzo What are your thoughts on this, should we wait and do everything in a new PR or should I add in a fix for the display stuff only and we can address the component name (or not) later?

keianhzo:
Not super concerned about the misspelled `frustrum` component name as long as the user facing text is correct so I'm ok with just fixing that for now.

keianhzo (approved):
LGTM

PR #133, opened 2022-09-19
Fix #111

### LINKED ISSUE #111: Baking reflection probes doesn't work after changing tmp_path in the preferences.
**Description**
Newly baked images aren't used after changing the `tmp_path` property in the preferences and rebaking.

**To Reproduce**
- Bake a reflection probe.
- Change `tmp_path` in the preferences to a different folder.
- Change something in the scene, e.g. the world background.
- Bake the reflection probe again.
- Check the image editor and see that the image hasn't been updated.

**Expected behavior**
The new image should have replaced the old one for the reflection probe.

**Hardware** 
 - Add-on Release Version: 1.0.0 beta (all of them I believe)
 - Blender Version: 2.93.9, 3.3.0 (probably everything in between)

Exairnous:
Fix baking reflection probes after changing the tmp_path by adding an additional check to see if the file path of the new bake is different from the file path of the current bake and update the image for any image editors that are currently displaying it (this mimics the behavior of reloading the image when the tmp_path hasn't been changed).

### DIFF
--- a/addons/io_hubs_addon/components/definitions/reflection_probe.py
+++ b/addons/io_hubs_addon/components/definitions/reflection_probe.py
@@ -196,9 +196,15 @@ class BakeProbeOperator(bpy.types.Operator):
                     img = bpy.data.images.get(image_name)
                     img_path = "%s/%s.hdr" % (get_addon_pref(context).tmp_path,
                                               probe.name)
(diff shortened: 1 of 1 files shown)


keianhzo (approved):
LGTM

PR #135, opened 2022-09-20
Initialize networked component with a unique network id

keianhzo:


Exairnous:
It looked like you were going to wait for Dom to review this, otherwise I would have mentioned this earlier.  The `init` method won't be triggered when an object is duplicated, so if you need the network id to be unique for each component, you'll need to trigger it after duplication.  I've done some testing and it looks like it might work to add a depsgraph_update_post handler that will trigger on the duplicate operators:

```python
@persistent
def depsgraph_update(dummy):
    wm = bpy.context.window_manager
    if wm.operators:
        if len(wm.operators) != wm.operators_length:
            if wm.operators[-1].bl_idname.startswith("OBJECT_OT_duplicate"):
                print(wm.operators[-1])
    wm.operators_length = len(wm.operators) # operators_length is an IntProperty added to bpy.types.WindowManager
```

PR #142, opened 2022-10-12
Add bloom to environment settings

netpro2k:
This adds a checkbox for 'HDR render pipeline" which allows configuring bloom settings. It should be considered experimental and checking this box will opt you into some future breaking scene changes that may require a re-export. (The first of which will likely be automatic application of PI on lightmap intensity when needed).

When this is enabled tonemapping mode will currently be ignored and ACES will always be used. You can still set an exposure, though its meaning may change at some point. This will likely become configurable again in the future but with different options (hopefully with adaptive exposure and custom LUT).

See https://github.com/mozilla/hubs/pull/5742 for more details

### DIFF
--- a/addons/io_hubs_addon/components/definitions/environment_settings.py
+++ b/addons/io_hubs_addon/components/definitions/environment_settings.py
@@ -1,4 +1,4 @@
-from bpy.props import FloatProperty, EnumProperty, FloatVectorProperty, PointerProperty
+from bpy.props import FloatProperty, EnumProperty, FloatVectorProperty, PointerProperty, BoolProperty
 from bpy.types import Image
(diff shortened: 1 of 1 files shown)


Exairnous:
I've just done a quick scan of the code so far, but it seems generally good.  One thing I noticed is that I think it might be better if settings like "enableHDRPipeline" and "enableBloom" are always included in the export so that importing them is easier.  I think the stuff under those bools is fine to leave out if false though (i.e. the enableHDRPipeline is always there, if it's false don't include anything after, but if it's true then include the enableBloom, and so on).

keianhzo (approved):
LGTM

PR #143, opened 2022-10-16
Append/Link Migration Handling, Individual Component Versioning, and Gizmo Stabilization (1/3)

Exairnous:
@keianhzo @netpro2k 
I initially started work on allowing local migration of components when they are appended/linked (which required the individual components to be versioned), but later on I found that the infrastructure I had built for that was needed for some gizmo fixes/improvements (so I started a second branch based off the first), and then I found that I needed some of the new gizmo changes to the underlying infrastructure for some fixes to linked component migrations and then things continued from there.  I tried to keep things as orderly as possible and have three separate, relatively independent branches/PRs, but that didn't end up being feasible, so while I've still decided to just upload them all as PRs, they are basically just one big PR.  So sorry for the mess, but things turned out to be far more finicky and involved than I expected.  To try to keep things relatively clear, I'll outline everything I've done in this PR and then just provide links to the other PRs and leave them blank. (2/3) #144 (3/3) #145

So, without further ado, here's what I've done:

- Created an undo stack handler via hooking into depsgraph updates and capturing the undo steps via redirecting the C stdout. This is the only way I found to handle appending/linking, plus it also allows a lot of flexibility for other things and helps to ensure gizmos are always updated correctly.  I've made it as stable as I can and have confirmed it to work on both Linux and Windows, although it does unfortunately change the stdout buffering policy on Windows (this seems to have a very minimal impact, I'm only aware of one or two things that actually use the C stdout in Blender, so basically the end result is that if a user manually prints the undo steps they will be delayed until the next depsgraph update). I have yet to confirm it works on Mac because I don't have a Mac to test with.
    - There are two cases where this handler doesn't work.
        1. If you disable Global Undo in the preferences (but disabling this disables most of Blender's undo system and prevents modifying operator properties in the redo panel, so I don't think many people are going to do that).
        2. If you enable Undo Legacy in the Experimental tab of the preferences (I don't think many people will do this either).
    - However, even if people do enable one of these, I don't think this is a problem because the handler simply won't do anything and there are fallback operators they can trigger to manually perform migrations or update gizmos. 

- Reworked the migration system to allow individual migrations (it also probably improves performance).

- Added versioning to individual components.

- Added a migration report that pops up when a migration is first performed.  Components provide custom reports that can be based on the type of migration (GLOBAL for file migration and LOCAL for anything else).  I wouldn't have added this at this stage, except that some previous beta components simply cannot be migrated correctly and people have to be notified of this.  The last report can be viewed again via the View Last Hubs Report menu item at the bottom of the Window menu, there is also an option for the user to open the report in the info editor.
    - One of the things I tried initially for the UI of the report viewer was using a UI List (which is scrollable), but that kept crashing Blender (I'm not exactly sure why, so it may be possible a way could be found to make it work), so I rolled my own interface, which I actually quite like (plus it's not something you should be skimming anyway).

- Added a Migrate Hubs Components menu item to the Objects menu so that users can specifically trigger migrations (as a fallback).

- Fixed a bug with Spawner components erroring out on migration (Apply Gravity was only saved to the blend file if the user had modified it).

- Fixed a bunch of corner cases where gizmos wouldn't be displayed and made sure gizmos are displayed properly in multiple windows.  I also changed it so the gizmos depsgraph handler is kept registered the entire time and removed the object name msgbus subscriptions because it turned out they were unneeded.

- Added a Refresh Hubs Gizmos button to the gizmos popover in the 3D View.

- Added a utility functions for looping through components on an object, another to provide consistent word wrapping so that long lines can be spread across multiple Blender ui labels to avoid truncation, and also one to get the add-on version as a tuple.

However, there arose several open questions (and bugs related to them) that need to be answered before this will be ready for a full review.

1. How should linked components be handled?

    - I discovered during my work on this that linked components are treated much differently than appended components.  Linked components are special in that they are read directly from other blend files and cannot be permanently modified unless made local.  While migrating them does work (and is exported correctly), it isn't permanent and has to be reapplied in certain circumstances.  I have accounted for most of these circumstances, but have not accounted for when a deletion is undone (it seems like a bad idea to me to perform a migration whenever a deletion is undone), and I was in the process of adding a warning when the whole file is initially migrated that I realized that migrations for linked components won't save and will have to be redone each time the file is opened.  At this point I realized that I needed more input before deciding on a course of action.
    
    - The two possible solutions I see to this problem are:
    
        1. Continue on with my first idea and automatically migrate the linked components in most cases, but display a warning that it's unstable (I was in the middle of refactoring when I decided to put this up, so don't worry too much about the implementation now as it will change).
        
        2. Skip migrating linked components and display a warning to the user telling them they have to manually migrate (i.e. save over) all the files they linked from before the linked components will display properly.

2. Should components be versioned with the add-on version or individual versions that are specific to the component?

    - @netpro2k and I were talking about this at a Friday meetup and felt it needed more discussion as there are benefits to both methods (e.g. individual versions are more desirable if we're looking to integrate into an outside ecosystem of tools, while using the add-on version is simpler if we're sticking to just the add-on)

There are also some smaller questions I had:

- During the course of my development of the C stdout redirect, I found that there is an additional way people handle getting the C Runtime on Windows when a debug version of Python is used.  As far as I can tell this isn't relevant to Blender, but I'm not very familiar with ctypes, so I don't know if it's a good idea to add it just in case?

- @keianhzo There is a call to the refresh method of the GizmoGroup at the bottom of the setup method, but from what I can tell it doesn't seem to be needed, was there a reason you added it, or can it be removed?

- I've added in some additional, non-panel related, ui elements to panels.py and I'm wondering if I should rename it to ui.py? (And also whether the report viewer should be moved to ui.py, it's currently in operators.py)

- I've used strings to specify the migration type so far, but they could instead be an Enum.  I'm not sure what's preferred here?

- HubsComponentsExtensionProperties.version is currently stored on the scene, so if you have multiple scenes there could be multiple versions recorded (although this won't have any effect on migrations), what are your thoughts on moving it to be a property of the window manager?  I don't know if this needs to happen now, but I thought I should bring it up

And I think that's everything.  Hopefully it's clear enough to get started with.

Oh, here are the links to the reported issues this addresses.
- Closes https://github.com/MozillaReality/hubs-blender-exporter/issues/118
- Closes https://github.com/MozillaReality/hubs-blender-exporter/issues/119
- Closes https://github.com/MozillaReality/hubs-blender-exporter/issues/120
- Closes https://github.com/MozillaReality/hubs-blender-exporter/issues/138


keianhzo:
I answer the questions above here while I finish the review:

**Migration**
- I think it’s acceptable to perform migration when the object is linked and every time the scene is loaded.
- Why is it a bad idea to perform a migration when a deletion is undone?
- Solutions: I’ll rather avoid making the user have to explicitly take actions as people have tendency to not read warnings

**Components versioning**
- I think we mentioned this in a previous meeting. Adding versioning to components sounds reasonable so I think. This will be also persisted when exporting a GLTF and used for migration when importing a GLTF.

**C stdout redirect**
Sounds to me like a corner case, I’m ok with not spending time with now.

**panels.py Vs ui.py**
I think ui makes more sense at this point.

**Strings Vs Enums**
I personally prefer using typed parameters rather than strings. It makes clearer what values are supported and you can use code completion.

**HubsComponentsExtensionProperties**
I agree that this should be stored higher in the hierarchy so WM sounds good to me.

Exairnous:
> **Migration**
> Why is it a bad idea to perform a migration when a deletion is undone?

Because doing a migration involves looping through all the objects/bones/scenes in the blend file and could be costly performance-wise, especially if the add-on is left enabled when someone is working on a large, non-Hubs related project.  I'll do some testing to see how costly doing a migration is on a somewhat large project, if it isn't noticeable then I'm fine with adding that case to the undo handler, if it is fairly costly then I'll add a warning about undoing deletion as discussed in the add-on meeting.

> Solutions: I’ll rather avoid making the user have to explicitly take actions as people have tendency to not read warnings

Good point, and I think warning the user that it's unstable and probably not a good idea, while also doing what we can to keep linked components migrated so that they can at least look around the blend file, is reasonable.  Perhaps it would be a good idea to include an explicit suggestion to migrate the other files first?

> **Components versioning**
> I think we mentioned this in a previous meeting. Adding versioning to components sounds reasonable so I think. This will be also persisted when exporting a GLTF and used for migration when importing a GLTF.

Yes I'm pretty sure we did.  As mentioned in the meetings, having component specific versions is required to support user-defined components.  So since it looks like everyone agrees, I'll switch things to per-component versions.

> **C stdout redirect**
> Sounds to me like a corner case, I’m ok with not spending time with now.

If you're talking about the debug python on Windows, then I'm good with that.  I'm still not sure if it's even a case at all.

> **panels.py Vs ui.py**
> I think ui makes more sense at this point.

Great, I'll rename it then.

> **Strings Vs Enums**
> I personally prefer using typed parameters rather than strings. It makes clearer what values are supported and you can use code completion.

Okay, Blender's API mostly exposes stuff like this with Strings but there are definite advantages to Enums (although they are longer), I'm fine either way, so I'll switch it to Enums.

> **HubsComponentsExtensionProperties**
> I agree that this should be stored higher in the hierarchy so WM sounds good to me.

Okay, I'll see about moving it.

keianhzo:
Closing this as https://github.com/MozillaReality/hubs-blender-exporter/pull/145 has been landed.

PR #144, opened 2022-10-16
Append/Link Migration Handling, Individual Component Versioning, and Gizmo Stabilization (2/3)

Exairnous:
Append/Link Migration Handling, Individual Component Versioning, and Gizmo Stabilization (1/3) #143
Append/Link Migration Handling, Individual Component Versioning, and Gizmo Stabilization (3/3) #145

PR #145, opened 2022-10-16
Append/Link Migration Handling, Individual Component Versioning, and Gizmo Stabilization (3/3)

Exairnous:
Append/Link Migration Handling, Individual Component Versioning, and Gizmo Stabilization (1/3) #143
Append/Link Migration Handling, Individual Component Versioning, and Gizmo Stabilization (2/3) #144


addons/io_hubs_addon/components/hubs_component.py
```diff
@@ -21,6 +23,11 @@ class HubsComponent(PropertyGroup):
         'icon': 'icon.png'
     }
 
+    # Properties defined here are for internal use and won't be displayed by default in components or exported.
+
+    # Version of the add-on this component was created with.
+    addon_version: IntVectorProperty(size=3)
```
keianhzo, line 29: The way I see this is `HubsComponent` having another definition for a version and a class method to get it, same as we are doing for the other definitions:

```python
    _definition = {
        # The name that will be used in the GLTF file MOZ_hubs_components object when exporting the component.
        'name': 'template',
        # Name to be used in the panels, if not set the component name will be used
        'display_name': 'Hubs Component Template',
        # Category that is shown in the "Add Component" menu
        'category': Category.MISC,
        # Node type to where the component will be registered
        'node_type': NodeType.NODE,
        # Panel types where this component will be shown
        'panel_type': [PanelType.OBJECT],
        # The dependencies of this component (by id). They will be added as a result of adding this component.
        'deps': [],
        # Name of the icon to load. It can be a image file in the icons directory or one of the Blender builtin icons id
        'icon': 'icon.png',
        # Version of the component. This will be used to trigger component migrations.
        'version': (0, 0, 1)
    }

    @classmethod
    def get_version(cls):
        return cls.__get_definition('version', (0, 0, 0))
```

and an instance property `version`:
```python
version: IntVectorProperty(size=3)
```

then in `handlers` I think we could get rid of the GLOBAL/LOCAL (not sure though as we are using it for the report in the migrate methods and that might be local specific?) parameter and just compare the class version against the component version and migrate the component whenever they differ:
```python
component.version < component.__class__.get_version()
```

and we can probably refactor some code there as most of the migration cases will look something like where only the migration host changes:
```python
    component_version = component.__class__.get_version()
    try:
        if version < component_version:
            migration_occurred |= True
            component.migrate(version, scene, migration_report)
            component.version = component_version
```
I think you mentioned that there is some case when we still need to compare against the global add-on version but I can't remember when was that.

addons/io_hubs_addon/components/utils.py
```diff
@@ -12,7 +18,9 @@ def add_component(obj, component_name):
 
     component_class = get_component_by_name(component_name)
     if component_class:
-        update_gizmos()
+        if 'create_gizmo' in component_class.__dict__:
```
keianhzo, line 21: I guess there is some case that I'm missing but as all components inherit from `HubsComponents` won't this always be true?

addons/io_hubs_addon/components/utils.py
```diff
@@ -112,3 +121,129 @@ def redraw_component_ui(context):
...
+    for component_item in host.hubs_component_list.items:
+        component_name = component_item.name
+        component_class = get_component_by_name(component_name)
+        if not component_class:
+            continue
+
+        component = getattr(host, component_class.get_id())
+        yield component
+
+def wrap_text(text, max_length=70):
```
keianhzo, line 223: This is great to have.I think we could extend the use of this to all widgets texts in the future. Blender doesn't wrap by default and some important things like warnings get truncated and are not easily readable.

addons/io_hubs_addon/components/utils.py
```diff
@@ -112,3 +121,129 @@ def redraw_component_ui(context):
...
+
+# Note: Set up stuff specifically for C FILE pointers so that they aren't truncated to 32 bits on 64 bit systems.
+class _FILE(ctypes.Structure):
+    """opaque C FILE type"""
+
+
+if platform.system() == "Windows":
+    try:
+        # Get stdio from the CRT Blender's using (currently ships with Blender)
+        libc = ctypes.windll.LoadLibrary('api-ms-win-crt-stdio-l1-1-0')
```
keianhzo, line 134: I wonder how easily this can break with Windows updates 🤔

addons/io_hubs_addon/components/utils.py
```diff
@@ -112,3 +121,129 @@ def redraw_component_ui(context):
...
+            def c_fflush():
+                pass
+
+    except: # The C runtime wasn't found.  This is unlikely to happen, but if it does then just skip flushing since Linux/Mac doesn't seem to strictly require a C-level flush to work.
+        print("Warning: Unable to find the C runtime.")
+        def c_fflush():
+            pass
+
+
+@contextmanager
```
keianhzo, line 186: Interesting, I didn't know this.

addons/io_hubs_addon/utils.py
```diff
@@ -1,6 +1,10 @@
 def get_addon_package():
     return __package__
 
+def get_version():
```
keianhzo, line 4: Maybe we don't need this anymore if we refactor the versioning code

keianhzo (commented):
I left a few comments, we can discuss whenever you want.

keianhzo:
As this PR is quite big and introduces quite a few formatting updates, I think it might be a good idea to add the PEP8 formatter and save the modified files so it gets aligned with the overall format.

Exairnous:
> This looks great. I think we have greatly improved the code stability and handled a bunch of edge cases. I left a few comments, we can discuss whenever you want.

Thanks!

> As this PR is quite big and introduces quite a few formatting updates, I think it might be a good idea to add the PEP8 formatter and save the modified files so they get aligned with the project format.

Yes, I'll format them with a PEP8 formatter (probably autopep8 or pylint) and commit the changes.  It seems like vscode is using autopep8 for this, but there's also a .pylintrc file in the main add-on folder, so I guess both are acceptable?  It might be nice to include something like Blender's "make format" so that anyone can easily format the whole repo from the command line.

Exairnous:
I've updated this PR to the latest master and added in the discussed things, as well as handled any other issues relating to migrations, versioning, etc. that came up during the development.  See the commits for details, but here are some highlights:

- Per component versions are now used exclusively.
- The report viewer has been improved.
- The warning messages that are displayed on migrations have been improved and expanded.
- The scene version property has been removed and the components in any files where it was present will be updated to have proper per-component versions.
- A migration is now triggered when the add-on is enabled in the middle of a session.
- Material components are now supported in migrations.
- The cases where scene/view layer actions don't trigger a depsgraph update have been documented.
- Handling for multiple windows has been improved.
- PEP 8 formatting from the new formatting script has been applied.

keianhzo (approved):
LGTM Great work @Exairnous!

PR #152, opened 2022-11-06
Change overrideAudioSettings to default to True.

Exairnous:
Having overrideAudioSettings default to false broke old files because previously the audio parameters were always applied.

This also brings up the question of how we want overrides to act in regard to the different components, e.g. audio zones will always need to override the audio settings, while videos will probably only override the audio settings occasionally.


keianhzo (approved):
I'd rather keep False as default but now it's tricky so we can look into it when we land the component versioning.

PR #154, opened 2022-11-07
Include File Name in Reflection Probes Baked Images.

Exairnous:
This PR includes the name of the blend file in the reflection probes' baked image file names.  This reduces name collisions when using the default relative path for
reflection probe bakes and there are multiple blend files in the same folder or when using a single absolute path for all bakes.

If the blend file hasn't been saved yet, "untitled" is used as the blend file name.

Name conflicts are still possible if blend files in different folders are named the same.

I considered using uuids to prevent name collisions, however these aren't human friendly and also either prevent overwriting previous images for the same probe or would be difficult to manage if a uuid was assigned to a probe and all of its bakes used that as the file name (it would face similar problems to what the networked component did).


Exairnous:
Closing this since #156 has been merged.

PR #156, opened 2022-11-07
Store Reflection Probe Bakes in Blend Files.

Exairnous:
Instead of baking the files to an output directory, this PR bakes them to Blender's session temp directory and then immediately packs and discards them.  This allows Blender to do all of the file management for us and will never be in danger of overwriting another blend file's reflection probe bakes.  Bakes of the same name will essentially be overwritten the same way they are now.  Baking to a new probe name still won't overwrite the previous bake, but the image will only hang around until the blend file has been saved and opened again (provided it has no users), so I consider that a small improvement.  Because we're now dealing with only Blender's internal "file system" I've had to account for a few things that we didn't have to when outputting to a permanent folder, but it's also allowed the opportunity to stabilize and improve a few things.

- I've managed to reduce the chance of accidental overwrites of the wrong probe when Blender juggles names to keep everything unique, and make it easier to see that this is what Blender has done.  This doesn't completely prevent the possibility of accidental overwrites, but to do that would require a naming scheme change and a migration, so this would need to happen later.  See the relevant commit for more details.
- I found the reason for that delay needed between renders and prevented it from occasionally deadlocking.
- I added export, import, and open functionality for interoperability with outside compositing/image editing applications.  When the bakes were output to the file system anything could easily use them, but now that they're packed, specific IO functionality was needed to accommodate this, plus it's nice to be able to just open any file and add it to a probe.
- I added the ability to lock probes, i.e. prevent their environment maps from being rebaked/changed.  This was needed to prevent accidental overwrites of linked/appended images when their names conflict with probes (these kinds of accidental overwrites weren't possible when the bakes were stored on the file system), and it's overall a nice feature to have.
- I fixed a case where the bake operator would still try to bake even though there weren't any valid probes (e.g. baking selected, but none of the selected objects were probes)
- I've added warnings for probes that don't match the specified resolution and provided an operator to select these.  Before the addition of locking just displaying the warning on resolution changes would do, but since baking would skip locked probes and you could end up with an invalid state but no warning, additional warnings needed to be added to accommodate this.
- I've added in a warning for image data-blocks that can't load their image, e.g. if a non-packed file had an invalid file path.
- I've added tooltips to the bake operator to clarify what gets baked, e.g. bake all only bakes all the probes in the current view layer.
- I prevented the saved state of the preferences from changing when baking, e.g. before this change baking would always mark the preferences as changed and in need of saving, when auto-save preferences was off, but now it will be put back to it's previous state.
- And I've added in support for linked probes, by which I mean notifying the user that linked probes can't be baked/changed (which keeps it consistent with the behavior of other linked components) and linked probes are skipped on bake/import/export.

I also removed the tmp_path preference as it wasn't needed any more when storing the bakes inside the blend file.

Overall I've found this approach to be much stabler/nicer, the only downside is that now each blend file will contain a copy of all the reflection probe environment maps, which increases their file size, but since reflection probe images are generally small I don't think it will bloat things too much.

Note: After packing, the stored file path is changed to be relative to the blend file so that it's simpler and more pleasant to use, however this causes an error to be printed to the terminal, but everything works fine so it seems it can just be ignored.  If anyone's really worried about it we can just not change the file path and leave it pointing to the old temp directory, which won't look as nice, but would prevent it printing erroneous errors to the terminal. (My vote goes to just ignoring the fake errors)

Alternatives I considered: #154 #155

### DIFF
--- a/addons/io_hubs_addon/components/definitions/reflection_probe.py
+++ b/addons/io_hubs_addon/components/definitions/reflection_probe.py
@@ -209,6 +209,16 @@ class BakeProbeOperator(bpy.types.Operator):
                     # Store the old image's full name in case of name juggling.
                     old_img_name_full = old_img.name_full if old_img else ""
 
+                    conflicting_img = None
+                    for img in bpy.data.images:
+                        if img.name == image_name and not img.library:
+                            conflicting_img = img
+                            break
+
+                    if conflicting_img and conflicting_img != old_img:
+                        # Rename the conflicting image to help avoid problems caused by Blender's name juggling and allow name juggled images to be more easily found.
+                        conflicting_img.name = f"{conflicting_img.name}-old"
+
                     img_path = get_probe_image_path(probe)
                     img = bpy.data.images.load(filepath=img_path)
                     img.name = image_name
--- a/addons/io_hubs_addon/components/definitions/reflection_probe.py
+++ b/addons/io_hubs_addon/components/definitions/reflection_probe.py
@@ -60,7 +60,7 @@ def set_resolution(self, value):
         self.resolution_id = RESOLUTION_ITEMS[value][0]
 
 
-def get_probes(all_objects=False):
+def get_probes(all_objects=False, include_locked=False):
     probes = []
     objects = bpy.data.objects if all_objects else bpy.context.view_layer.objects
     for ob in objects:
@@ -73,6 +73,9 @@ def get_probes(all_objects=False):
                 component_name = component_item.name
                 if component_name in registered_hubs_components:
                     if component_name == 'reflection-probe':
+                        probe_component = ob.hubs_component_reflection_probe
+                        if probe_component.locked and not include_locked:
+                            continue
                         probes.append(ob)
 
     return probes
@@ -444,7 +447,7 @@ class ImportReflectionProbeEnvMaps(Operator):
 
             if not imported_file:
                 num_failed += 1
-                self.report({'WARNING'}, f"Warning: Couldn't import {f.name}.  The corresponding probe doesn't exist")
+                self.report({'WARNING'}, f"Warning: Couldn't import {f.name}.  The corresponding probe either doesn't exist or is locked.")
 
 
         if num_failed:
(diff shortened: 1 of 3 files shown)


keianhzo (commented):
Thanks for this, this fixes a bunch of bugs, improves the usability and verbosity and add some new cool features to reflection probes.

I'd like to discuss further the reflection probe output and all the implications that it has. I think Blender users are used to have render results stored in memory and be responsible for persisting them in disk (which might or might not be what they want depending on the case). On top of being the expected behavior this saves us from dealing with multiple potential corner cases (ie. if they want to override it's their choice, we just server the output) and users would assimilate this behavior naturally. We should have probably done this since the beginning but it would be nice to discuss what route do we want to take, for this and similar future cases, before releasing.

Exairnous:
I think the route has been decided on after the discussion in discord (starting here: https://discord.com/channels/498741086295031808/964183314674098276/1045319656556208160).

So I've updated this to the latest master, and I think it can be merged whenever you're happy with it.

I decided to add in the disabling of the operators for linked reflection probe components that was talked about in #157 here because it was easier/made more sense (e.g. the Open EnvMap operator should also be disabled, but it's only present here).  I managed to add a disabled message to the tooltip, but 2.93 doesn't get the nice red text.

I thought about doing the same for when reflection probes are locked, but decided against it because it didn't really seem necessary/much of an improvement (I thought the red text was a bit too strong and 2.93 doesn't get the red text, so it's inconsistent, plus most of the time the locked icon will be clearly visible/highlighted, unless the operators are added to the quick favorites menu, but then it's not going to be consistent anyway).  I did add in a check for locked probes to the Open EnvMap operator's poll, though, to prevent any possibility of it modifying a locked probe.

Exairnous:
Also, on a separate note, I ran the new format script on the branch before pushing, however the CI test is failing on the formatting, so I guess one of them needs some adjustment.

keianhzo:
Yeah the `formatter` deals with a few things but doesn't really format everything and like breaks is one of them. I've added a `pycodestyle` check exception to ignore the long lines warnings. In this case we just need to make this:

```python
    is_selected = (context.selected_objects ==
                   mismatched_probes and not context.active_object)
```

A single line and `pycodestyle` will be happy. If it's annoying we can also disable in the future.

I've fixed a conflict after merging my extras PR.

PR #157, opened 2022-11-07
Support linked Pointer Properties in Blender 3.2+

Exairnous:
This PR properly disables pointer properties when they're part of linked components in Blender 3.2+ and also adds a disabled indicator button to the left of the pointer property field to inform the user whether the selected data-block is linked or a link_override (via differing icons) and displays the source library name/path in a tooltip.  Note: The link indicator button is needed for Blender versions before 3.2 as well because Blender doesn't always show the source library in its default tooltips.

To add the link indicator I created a generic tooltip label operator that functions similarly to a regular label, but will open a custom tooltip when hovered.  Note: this can be used anywhere a label can be and I think it'll probably be useful in other places for warnings that are currently being truncated, but would look too busy wrapped.

There are several intricacies when it comes to Blender displaying tooltips and how I have the tooltip label set up:
- If the text of the tooltip label is being truncated in the UI then the first line of the tooltip will show the text without truncation.
- If the tooltip label isn't being truncated, then the first line will contain three dashes, provided the tooltip has custom text, otherwise the first line will be blank.
- If there is no custom text, the tooltip will display a single period in the second line (Blender automatically adds one by default).
- Multi-line tooltips are a thing, just add new line characters to your string. (But colors and/or formatting sadly aren't)

Alternatives I considered:
Using the tooltip label to display the library name underneath the pointer property with the full path displayed in the tooltip.  However, this made the component take up more space and was visually noisier.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/audio_target.py
+++ b/addons/io_hubs_addon/components/definitions/audio_target.py
@@ -1,8 +1,9 @@
 from email.policy import default
 from bpy.props import FloatProperty, BoolProperty, PointerProperty, EnumProperty, StringProperty
 from ..hubs_component import HubsComponent
-from ..utils import has_component
+from ..utils import has_component, is_linked
 from ..types import Category, PanelType, NodeType
+from ..ui import add_link_indicator
 from bpy.types import Object
 from ...io.utils import gather_joint_property, gather_node_property, delayed_gather
 from .audio_source import AudioSource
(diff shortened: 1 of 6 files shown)


keianhzo:
Blender already shows a data-block level linked icon that you can also use for  converting in a local data-block or overriding it. I think this PR does the right thing disabling the property when it belongs to a linked data-block but I'd like to understand better why we need an specific custom linked button/tooltip for this as Blender already shows one that users are familiar with and there is information about the linked blend file pat if you hover the data-block. Maybe I'm missing cases.

As you mention the tooltip can be useful in some cases were we might be trying to show too much information to the user.

keianhzo:
This has been clarified and this is covering some cases not covered natively by the Blender UI like when a reflection probe envMap pointer property is referencing a linked image from another blend file so we decided to keep up as it's here.

One thing that was agreed to be added is disabling the Bake button for linked reflection probe components as the linked image cannot be updated. If possible the tooltip should inform that the baking is not allowed in library linked components.

Exairnous:
I've rebased this on the current master.  I decided to disable the operators on linked probes in #156 because it was easier/made more sense (see the comment I left there for more details).  However, I think this should be merged in after #156 because it'll apply easier/cleaner on top of it than the other way around.  So I'll rebase this again once #156 is merged, but then I think it can be merged if you're happy with it.

keianhzo:
That shound like a plan.

Exairnous:
Okay, I think this should be good to go now.

keianhzo (approved):
LGTM

PR #158, opened 2022-11-09
Add rotation and scene envmap warnings to reflection probes.

Exairnous:
Add a warning about the scene not having an environment settings component with an environment map to both the scene and object reflection probe panels. Add a warning to the object reflection probe panel about rotation being applied to the probe.


keianhzo (approved):
Usability improvements for the win! Thanks!

PR #159, opened 2022-11-10
Rename panels.py to ui.py

Exairnous:
This will give us an official, central, and easily recognizable place to add generic UI elements and UI hooks (i.e. additions to Blender's predefined menus, etc.) as they are developed.


keianhzo (approved):
LGTM. Will land when the conflict is solved.

PR #167, opened 2022-11-29
Video-texture-target load slowdown

keianhzo:
This PR tries to address the video-texture-target import slowdown by improving the modules import process to only register the class that is contained in the currently iterated module.

I think though that there is something else going on with this slowdown that I couldn't really find.

### DIFF
--- a/addons/io_hubs_addon/components/components_registry.py
+++ b/addons/io_hubs_addon/components/components_registry.py
@@ -9,6 +9,7 @@ import inspect
 import os
 from os import listdir
 from os.path import join, isfile, isdir, dirname, realpath
+import pyclbr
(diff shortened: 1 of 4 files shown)


Exairnous:
This is weird.  I can no longer reliably reproduce the slowdown on either master or this branch.  So, I don't know what the issue is.

Looks like this is a good improvement to the component registration process, though.

I think this treatment could be applied to the audio-target component in this PR as well, and it looks like the `pyclbr` import is unneeded.

keianhzo:
Ops yes, added AudioTarget too and removed the unused import.

Exairnous:
LGTM.

PR #169, opened 2022-11-30
Limit Cone Outer Gain to the 0-1 range.

Exairnous:
Hubs produces this error when trying to upload anything with a Cone Outer Gain higher than one: DOMException: PannerNode.coneOuterGain setter: <number> is not in the range [0, 1]

I happened across this while testing #167

Edit: I don't think this should require any kind of migration because values above one fail in Hubs.


keianhzo (approved):
Good catch!

Exairnous:
Thanks :)

PR #171, opened 2022-12-05
Fix gather_joint_property error in Blender 3.2+ when getting the vnode.

### LINKED ISSUE #170: Bone source nodes fail to export in Blender 3.2+
**Description**
Components with source nodes that are set to bones fail to export in Blender 3.2+ because joint isn't an attribute of vnodes in Blender 3.2+. 

**To Reproduce**
- Open Blender 3.2+.
- Add an armature.
- Add an audio source component to the armature's bone.
- Add an audio target component to the default cube.
- Set the audio target component's source field to point to the armature, then pick the bone with the audio source component from the bone dropdown menu underneath.
- Export the audio target and bone source to a gltf file.
- Check the exported gltf file and see that the audio target component wasn't exported or see that this error has been printed to the blender console:
    - gather_scene_hook fails on <io_hubs_addon.io.gltf_exporter.glTF2ExportUserExtension object at 0x7ff3ad99c0d0>
'VExportNode' object has no attribute 'joint'

**Expected behavior**
Components that point to bone sources should export from all supported versions of Blender.

**Hardware** 
 - Add-on Release Version: 1.0.0 beta 1 - 1.0.0 beta 7
 - Blender Version: 3.2+

Exairnous:
Fixes #170

When looking for the equivalent vnode for the bone being referenced, check the vnode's blender_bone, rather than the joint attribute which isn't present in Blender 3.2+.

While working on this, I noticed that the code to get the vnode in both `gather_joint_property` and `gather_node_property` would most likely fail if the vnode wasn't found, so I think it should either be changed to use `vtree.nodes.get(...)`, so that it'll safely fall back to None, or the vnode should be assumed to be present and the None fallback removed in `next(...)` and then just do `node = vnode.node` and forget the whole `or gather` part.

### DIFF
--- a/addons/io_hubs_addon/io/utils.py
+++ b/addons/io_hubs_addon/io/utils.py
@@ -232,7 +232,7 @@ def gather_joint_property(export_settings, blender_object, target, property_name
         else:
             vtree = export_settings['vtree']
             vnode = vtree.nodes[next((uuid for uuid in vtree.nodes if (
-                vtree.nodes[uuid].joint == joint)), None)]
+                vtree.nodes[uuid].blender_bone == joint)), None)]
             node = vnode.node or gltf2_blender_gather_joints.gather_joint_vnode(
                 vnode,
                 export_settings


keianhzo (approved):
LGTM. Thanks!

PR #173, opened 2022-12-05
Reset the bone EnumProperty when the srcNode PointerProperty changes.

### LINKED ISSUE #172: The previously selected bone is still targeted after switching objects in the source field.
**Description**
When switching objects pointed to in the source field for audio and video texture target components, the bone enum property is left on it's current value.  This could cause problems during export if the source field was changed to point to a non-armature object without clearing the bone enum property first because it would still be assumed that a bone is being exported.  Fixing a leftover bone target on a non-armature object is difficult because the bone enum property is no longer shown (the workaround to fix this is to first select an armature, then clear the bone enum property, and then switch back to the non-armature object).

**To Reproduce**
- Open Blender.
- Add an armature.
- Add an audio source component to the armature's bone.
- Add a mesh object, e.g. a plane.
- Add an audio source component to the plane.
- Add an audio target component to the default cube.
- Set the audio target component's source field to point to the armature, then pick the bone with the audio source component from the bone dropdown menu underneath.
- Switch the audio target component's source field to point to the plane.
- See the warning message about the bone not having an audio source component, even though the plane is selected and the bone enum property is hidden.
- Export the scene to a gltf file.
- Check the exported gltf file and see that the audio target component wasn't exported or see that this error has been printed to the blender console:
    - gather_scene_hook fails on <io_hubs_addon.io.gltf_exporter.glTF2ExportUserExtension object at 0x7f7fdc50cd60>
'NoneType' object has no attribute 'bones'

**Expected behavior**
The bone enum property should not be applied to non-armature objects

**Hardware** 
 - Add-on Release Version: 1.0.0 beta 1 - 1.0.0 beta 7
 - Blender Version: 2.93.9-3.3.1

Exairnous:
Fixes #172

This also improves usability because I think having the bone be reset when choosing an object is more in line with what a user would expect.

I considered checking whether the object was an armature and only gathering bones if that was the case, but decided against that route because it is more complex, not as future proof, and wouldn't provide the usability improvement.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/audio_target.py
+++ b/addons/io_hubs_addon/components/definitions/audio_target.py
@@ -86,7 +86,8 @@ class AudioTarget(HubsComponent):
         name="Source",
         description="The object with an audio-source component to pull audio from",
         type=Object,
-        poll=filter_on_component
+        poll=filter_on_component,
+        update=lambda self, context: setattr(self, 'bone', BLANK_ID)
     )
 
     bone: EnumProperty(
--- a/addons/io_hubs_addon/components/definitions/video_texture_target.py
+++ b/addons/io_hubs_addon/components/definitions/video_texture_target.py
@@ -91,7 +91,8 @@ class VideoTextureTarget(HubsComponent):
         name="Source",
         description="The object with a video-texture-source component to pull video from",
         type=Object,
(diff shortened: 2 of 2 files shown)


keianhzo (approved):
LGTM Good catch, thank you.

PR #175, opened 2022-12-07
Fix error when clearing the srcNode property in audio and video texture target components.

Exairnous:
Clearing the srcNode property triggers a reset of the bone EnumProperty, but this failed because of an unguarded check for the mode of the srcNode object when accessing the bone EnumProperty to reset it.  This PR adds a check for whether the srcNode has an object before checking whether the srcNode object is in edit mode so that resetting the bone EnumProperty to the blank id doesn't cause an error when getting the list of bones.


keianhzo (approved):
LGTM

Issue #126, opened 2022-09-15
Pre/post export hooks don't loop through all objects.


Exairnous:
**Description**
The pre/post export hooks currently loop through the current view layer's objects instead of all the objects in the blend file.  This can cause problems with export because the glTF exporter doesn't use view layers as a determining factor when exporting objects and may also be exporting multiple scenes, so some of the objects exported may not have the hooks applied.

**Expected behavior**
The pre/post export hooks should loop through all the objects in the blend file.

**Hardware** 
 - Add-on Release Version: 1.0.0 beta 4
 - Blender Version: N/A

keianhzo:
Already addressed

PR #184, opened 2022-12-11
Make the pre/post export hooks loop through all the possible component hosts.

Exairnous:
Depends on #145
This handles the pre/post export hooks in a manner similar to the new migrations.
Fixes issue #126

The relevant commit for this PR is the last one listed.


keianhzo (approved):
LGTM

PR #185, opened 2022-12-11
Improve/fix the loop animation message bus handling and handler registration.

Exairnous:
This condenses and cleans up the message handling the same as is done in gizmos.py.
Fixes the the undo/redo handler function being appended/removed to/from the load_post handler instead of the proper undo and redo handlers.


keianhzo (approved):
LGTM

PR #186, opened 2022-12-13
Fix 3.2 classloading issues

keianhzo:
This is a similar case to https://github.com/MozillaReality/hubs-blender-exporter/pull/166#issuecomment-1335083605
We were importing some classes when the add-on is loaded but the glTF add-on is not yet loaded. The glTF add-on does a lot of `isinstance` comparisons and some of them were failing as the class instance that we have vs what the add-on has differ. 

Going forward we need to keep in mind to lazy load any class that we are using from the glTF add-on namespace.


Exairnous:
I did a little digging and found out some more about the issue.  It's present in both 2.93 and 3.4.  It seems that Blender loads add-ons in the order they're enabled, so by default the glTF one is loaded first and the Hubs one is loaded after and everything is fine, but if the glTF add-on is disabled and re-enabled after the Hubs add-on, then the issue appears .  I doubt many people are turning these add-ons off and on again, which is probably why it hasn't been noticed until now.

Anyway, I've tested it and everything seems fine with this PR applied, so LGTM.

keianhzo:
When an addon is enabled Blender checks if the module has been already loaded, in case it is, it forces a module reload: 
https://github.com/blender/blender/blob/c82311bb8efd24abec6f50a18256c636b78ef626/release/scripts/modules/addon_utils.py#L259

We had some references to the `io_scene_gltf2` module and that triggered a module load when the Hubs addon was loaded. In case the Hubs addon was loaded before the `io_scene_gltf2` addon, this would mean that Blender would force a `io_scene_gltf2` addon reload and any references we had to that addon class instances would no longer match the `io_scene_gltf2` ones and random exceptions would happen.

This PR defers any `io_scene_gltf2` modules references so they are loaded at export time so we make sure that we are pulling the right class references.

Note: The addon loading order is based on the order the addons were enabled. As the io_scene_gltf2 addon is a builtin addon, it's usually loaded before any other post-install enabled addon unless it has been disabled/enabled afterwards.

PR #187, opened 2022-12-14
Additional 3D gizmos for Hubs components

theanine3D:
I've created 7 additional 3D gizmos for the following Hubs components:

- audio
- video
- image
- spot light
- directional light
- point light
- link

These components will now show a visual indicator when you add these components to an Empty in Blender.
image
Untitled

The models are all very lightweight, all under 300 triangles, and all have a unique silhouette which makes them easy to identify.
Note: the 3 light gizmos all change color in real-time based on the color currently designated in their Hubs component, thanks to @Exairnous's addition below.


Exairnous:
Hey, this is wonderful!  It's great to have new gizmos!

I see you've addressed pretty much all the stuff we talked about at Taco Tuesday Testing Day, so thank you for that, I like it!

However, I found a couple code related things that need to be addressed:
- The spot light component doesn't need/shouldn't have a migrate function (and it's causing a Python error because migrate_networked hasn't been imported).
- The update_gizmo method is missing from the spot light and image gizmos.

And here are a few more design notes/questions to be considered:
- The point light gizmo still isn't centered.
- I've thought about it a bit more, and I think gizmos that extend outward should have their base at the center point, rather than centering the model on the center point, this would apply to the audio and spot light gizmos and I think it would make them easier to work with.
    - Also, possibly the audio component should have a 1m plane at the center point so that its positioning, e.g. whether its in-Hubs representation will be hidden behind objects, can be more easily detected?
- The image/video gizmos are 3d, but their Hubs counterparts are 2d.  While I like the 3d gizmos, perhaps it would be better to make them 2d (or at least slim them down) so that they more accurately represent what will show up in Hubs?
- The light gizmos could take on the light color (this would require a little finagling to be performant, but could be cool).
    - Test code for anyone interested (not optimized): replace the current `gizmo.color` line in `create_gizmo` with `gizmo.color = getattr(ob, cls.get_id()).color[:3]` and add this argument to the color property `update=lambda x, y: update_gizmos()`
- Should the light gizmos be drawn in screen space and maintain their size the same as the Blender lights do?  (I know an early version of the waypoint gizmo did this, but I don't remember how, off the top of my head)

Aside from the design and coding for this PR, the maintainability and extensibility of gizmo models should be addressed.  Currently, gizmo models have been built, converted to the requisite format, and added to the add-on, but the original models aren't kept with the add-on and there is no `import_gizmo` script to retrieve them from the Python model files.  Ideally, I think the blend files for the gizmo models should be kept in a separate folder within the repository, but there should at least be some way to edit the models, re-export them to a new format, or use them as a base for new gizmos, if needed.
This also brings up the question of what license they should be under (as far as I know, the MPL isn't really used for this kind of thing, but since they're intended only for use within the add-on, perhaps it makes sense?)
This is mostly directed at the Hubs team, but if anyone else has an opinion, feel free to chime in too.  This doesn't necessarily have to be addressed for this PR, but I think it should probably be addressed sooner, rather than later.

theanine3D:
I made the requested changes and uploaded them. The Image and Video gizmos are now mostly flat, which will make them more practical for previewing how they'll actually look in Hubs. The Audio component now has a flat plane at its center, to make positioning more practical as well. The Point Light is centered, though it's based on the bulb part of the model (which is where light would actually come out of), not based on the connector at the bottom. Let me know if that's not what you meant.

I also added your test code for the light entities showing their color - however, the gizmos don't appear to update their color automatically with this code, and instead only update themselves randomly afterward or if you rename the Blender object (which for some reason, forces an update of the color). This'll need fixing, but I'm unsure of how to fix it.

image

Exairnous:
Thanks.

> The Point Light is centered, though it's based on the bulb part of the model (which is where light would actually come out of), not based on the connector at the bottom. Let me know if that's not what you meant.

Yup, that's exactly what I meant.  Looking good.

> I also added your test code for the light entities showing their color - however, the gizmos don't appear to update their color automatically with this code, and instead only update themselves randomly afterward or if you rename the Blender object (which for some reason, forces an update of the color). This'll need fixing, but I'm unsure of how to fix it.

Ah, you need to import the `update_gizmos` function from `gizmos.py`, so you'll need to add it to the `from ..gizmos import` line everywhere you want to use it.  If you start blender from a terminal you'll see errors about it being undefined.  I guess since it will only update when using gizmos it won't be too much of a performance problem right now and it can be optimized later if need be.  In which case, the lambda arguments (`x` and `y`) should technically be `self` and `context` to keep it consistent with Blender's standard naming conventions.

All in all, it's looking good.  I like it, at least :)

theanine3D:
> Ah, you need to import the `update_gizmos` function from `gizmos.py`, so you'll need to add it to the `from ..gizmos import` line everywhere you want to use it. If you start blender from a terminal you'll see errors about it being undefined. I guess since it will only update when using gizmos it won't be too much of a performance problem right now and it can be optimized later if need be. In which case, the lambda arguments (`x` and `y`) should technically be `self` and `context` to keep it consistent with Blender's standard naming conventions.

This is fixed now in my latest commit. I tested it, and the light gizmos do update in real-time now, as they should. Thanks so much for the help!

Exairnous:
> This is fixed now in my latest commit. I tested it, and the light gizmos do update in real-time now, as they should. Thanks so much for the help!

Excellent!  You're welcome.

> I also modified the spot light further to make it more obvious that it is a light. It's been nagging at me that it looked too much like an audio speaker. There is now a light bulb in the middle of the spotlight to make it more distinct from a speaker.

Looks good.  I like it.

I found the code that'll draw the gizmos in screen space, you have to set `gizmo.use_draw_scale` to `True` and then you'll need to override the scale in the update_gizmo method (you can use the media frame as a reference for that).  It looks like a constant scale of (0.5, 0.5, 0.5) makes them close to the size of the Blender lights.

While it kinda makes sense to draw them at a constant size on the screen (since scaling doesn't affect the components anyway), it also seems like it could make things harder/busier with smaller (or larger) than normal scenes (or if you're just zoomed out to look at the big picture), and since scaling doesn't affect the end result in Hubs, incorporating the object scale like all the other gizmos may be nicer to work with, but I don't know.  Anyway, I'm interested in your thoughts on this after you try it out.

theanine3D:
I tried it out with the screenspace scaling.

My thoughts: Although it does make it easier to see the light gizmos easier anywhere they might be in the scene, they also obscure the scene itself too much. They become way too flashy and distracting.

Unlike Blender's own built-in lights (which are entirely greyscale and very uninstrusive in their look/appearance), these Hubs gizmos I've created have color and and are only _slightly_ transparent, making them much more intrusive in the 3D viewport when viewed with _use_draw_scale_ set to True.

Blender's light objects are able to look fine in screenspace because their design was kept incredibly simple - ie. tiny, drawn entirely with just dotted lines, no color.

For that reason, I believe sticking to object scale for the lights would be best. At least by default. Perhaps a toggle button could be added in the addon settings, so people can use draw scale instead of object scale if they really prefer.

Exairnous:
That's kinda the way I was leaning.  Having them scaled to 0.5 did help, I think, but it still made the scene busy (if you didn't bother modifying the update_gizmo method you can just scale the object by 0.5, BTW).  You also make a good point about differentiating them from the Blender lights.  So I think it's good to just leave them as they are without screenspace scaling, and as to having a preference, we can, although I think I'd leave that for another PR.

keianhzo:
I've added a commit to update this branch to the latest master code.

theanine3D:
@keianhzo Thanks! I zipped the blend and added the ZIP below. Unfortunately I'm not familiar with Git yet. 😅 

[gizmos.zip](https://github.com/MozillaReality/hubs-blender-exporter/files/10296562/gizmos.zip)

keianhzo (approved):
LGTM

PR #182, opened 2022-12-11
Handle components on unsupported hosts.

Exairnous:
Depends on #145
Displays a warning message about unsupported hosts in the migration report.
Adds a simple ui for the component with a warning and a remove button, the same as unknown components do.
Improves host checking for warning messages and deduplicates this.
Removes the unused OBJECT_DATA member from the PanelType Enum.

The relevant commits for this PR are the last five listed.

PR #188, opened 2022-12-18
Armature gizmo and migration fixes.

Exairnous:
Depends on #182

This PR migrates armatures separately from objects and refreshes gizmos when switching an object's armature.
Armatures exist independently from objects and don't have to belong to one, so they need to be migrated independently as well.  Since this is the first time armatures not attached to objects would be migrated, warnings will be shown on any migrations of unattached armatures, and they aren't versioned as migrated from beta files either.  Bones are also now updated from edit mode so that none are missed, and the messages have been improved to specify whether it's a regular object that has the component or an unattached armature.  As far as I know, unattached armatures aren't exported, so migrations should be the only place that needs to account for them at this time.

Note: the gizmo refreshing isn't perfect, the user may still have to manually refresh the gizmos, but it's at least better and there are no longer any Python errors.

The relevant commit for this PR is the last one listed.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/audio_params.py
+++ b/addons/io_hubs_addon/components/definitions/audio_params.py
@@ -114,7 +114,7 @@ class AudioParams(HubsComponent):
             self.coneOuterAngle = radians(
                 self.coneOuterAngle)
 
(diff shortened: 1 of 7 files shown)


keianhzo (approved):
LGTM

PR #192, opened 2023-01-12
Fix for Windows stdout issues

keianhzo:
We are interfacing with the C level stdout str to be able to read the ouput of `bpy.context.window_manager.print_undo_steps()` to find if an undo happened. For some reason that I don't really understand piping that stream seems to misbehave in Windows when the buffer is over 4k. Ideally we should find the reason for this but in the meantime I've workarounded this by replacing piping with a temp file. This is Windows only as undo steps are configurable (and some people might have quite high numbers) and we should avoid file I/O.

An alternative to this would be to access the `WindowManager`native instance through a ctype wrapper and get the `undo_stack` from there: https://github.com/blender/blender/blob/master/source/blender/makesdna/DNA_windowmanager_types.h but it will make us dependent on the internal `WindowManager` struct update which might be tricky to maintain (python mirror struct needs to be aligned with the actual struct members). For now, unless we can find out the reason for the behavior above, I think this will suffice.

### DIFF
--- a/addons/io_hubs_addon/components/__init__.py
+++ b/addons/io_hubs_addon/components/__init__.py
@@ -1,7 +1,8 @@
-from . import (handlers, gizmos, components_registry, ui, operators)
+from . import (handlers, gizmos, components_registry, ui, operators, utils)
 
 
 def register():
+    utils.register()
     handlers.register()
     gizmos.register()
     components_registry.register()
@@ -15,3 +16,4 @@ def unregister():
     components_registry.unregister()
     gizmos.unregister()
     handlers.unregister()
+    utils.unregister()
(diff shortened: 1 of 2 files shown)


Exairnous:
The new stuff looks generally good, but you need to seek back to the beginning of the file after truncation otherwise no data is read into Blender.

I haven't profiled, but it seems like this might be just as performant or more performant as using pipes, so that plus the fact that it would simplify and remove code duplication, and prevent any chance of us ever running into this problem again, makes me think that we might want to just use the temp file for all platforms.

keianhzo:
Yeah, I don't see much difference either so it might make sense to consolidate that in file access, seems to be more reliable.

PR #193, opened 2023-01-16
Merge-ci-steps

keianhzo:
Merge all CI steps in one job


Exairnous:
This generally seems fine, but it resets the job number that's copied into the version and could cause confusion when updating and conflicts with previous versions (i.e. two versions could end up with the same version string).

keianhzo:
Good catch. We are not using that last number for anything really other than knowing what version you are on but I can add 100 to fill that gap and make sure that we stay ahead.

Exairnous:
Thanks.  Actually, I think knowing which version you're on is fairly important (IMO).  Adding 100 will work, but I'm a little worried that it'll make figuring out which commit the add-on is based off of even harder.

Exairnous:
If that works that sounds good.

PR #194, opened 2023-01-26
New operator that generates navmesh for selected objects

theanine3D:
This PR adds a new operator ("Generate Hubs Navigation Mesh") to the addon. The operator is accessible via the 3D viewport's "Object" menu while a mesh object is selected in object mode.

The operator is pretty straightforward - it makes a bmesh copy of the original object, checks the normals of all the faces, and then removes any that are too steep to walk on. Then it cleans up the result a bit (limited dissolve, triangulate, remove loose verts). The result is a lower-poly, more optimized mesh than the original model.

The operator is fast and creates an instant result to work with. In many cases the result needs only minimal editing afterward. The operator also adds the navmesh Hubs component to the newly generated object as well.

navmesh-gen-1

It currently has one limitation - it currently fails at generating _connected_ staircase slopes, because the operator removes any steep faces, including the 90 degree vertical geometry in stairs. It will correctly grab the top (flat) face of each stair, but it won't connect them, so the user will still need to edit the navmesh afterward in edge cases like that. See GIF below. However, it only takes a few seconds of manual editing to turn these unconnected faces into a usable navmesh slope.

stair-example

At the moment, the operator also adds a new material to the resulting navmesh object (if the material doesn't already exist), with a bright neon pink color, to make it easy to distinguish the navmesh in the 3D viewport from the original object. I added this functionality because otherwise it's too easy to accidentally select the original object when you're trying to edit your navmesh afterward. I tried to pick a color that is less likely to already be present in most scenes. Because navmeshes aren't actually rendered once exported to Hubs, the navmesh material is just there to make editing in Blender easier, with hopefully no impact to performance.


addons/io_hubs_addon/components/ui.py
```diff
@@ -188,6 +188,7 @@ def object_menu_addition(self, context):
     layout = self.layout
     layout.separator()
     op = layout.operator("wm.migrate_hubs_components")
+    op = layout.operator("object.generate_hubs_navmesh")
     op.is_registration = False
```
keianhzo, line 192: I believe this setting belongs to the previous operator.

addons/io_hubs_addon/components/operators.py
```diff
@@ -433,6 +442,79 @@ def invoke(self, context, event):
...
+
+        # Make sure there's a mesh object set as active. If not, abort
+        if obj is not None:
+            if obj.type != "MESH":
+                return {'FINISHED'}
+        else:
+            return {'FINISHED'}
+
+        bm = bmesh.new()
+        bm.from_mesh(obj.data)
```
keianhzo, line 463: You probably need to apply transforms here so the actual geometry is transformed. Something like:
```python
        bpy.ops.object.transform_apply(location=False, rotation=True, scale=True)
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -433,6 +442,79 @@ def invoke(self, context, event):
...
 
+class GenerateNavMesh(Operator):
+    bl_idname = "object.generate_hubs_navmesh"
+    bl_label = "Generate Hubs Navigation Mesh"
+    bl_options = {'REGISTER', 'UNDO'}
+
+    def execute(self, context):
+        import bmesh
+
+        obj = bpy.context.active_object
```
keianhzo, line 453: Would be great if we can generate a nav mesh from multiple selected objects instead of only the active one

keianhzo (changes requested):
This looks really good! Thanks for the contribution. I have left a couple of comments. We can keep the discussion open here to refine this.

addons/io_hubs_addon/components/operators.py
keianhzo: Maybe also applying operators could be a good thing to do as the user probably expects the nav-mesh generation to happen on the resulting mesh. nav-mesh generation for surfaces?

theanine3D:
I've implemented all of the above requested changes/suggestions. However, the solidify / remesh combination has issues.

I added code for it but my tests with the Remesh modifier did not produce satisfactory results. So I left that part of the code commented. You can uncomment lines 493-495, 499-502, and 506-512, if you'd like to play around with it and see if maybe you can get a better result. My tests kept resulting in missing faces, large gaps in teh resulting model, large chunks of stairs still missing, etc.

All the other code changes work fine though. The updated code now supports generating 1 navmesh from multiple selected objects. The transforms are applied beforehand as well, and if any objects are multi-user objects, they're forced to become single-user before transforms are applied.

keianhzo:
Humm I didn't know that that addon already existed, I missed that in my search. but this works pretty good and it's not hard to set up. Might be good to give this a try to provide a fully functional solution.

PR #197, opened 2023-01-31
Add copy components support

keianhzo:
This PR adds copy properties support, a functionality that we lost in the add-on update.

### DIFF
--- a/addons/io_hubs_addon/components/operators.py
+++ b/addons/io_hubs_addon/components/operators.py
@@ -445,6 +445,12 @@ class CopyHubsComponent(Operator):
     panel_type: StringProperty(name="panel_type")
     component_name: StringProperty(name="component_name")
 
+    @classmethod
+    def poll(cls, context):
+        panel = getattr(context, 'panel')
+        panel_type = PanelType(panel.bl_context)
+        return panel_type != PanelType.SCENE
+
     def get_selected_bones(self, context):
         selected_bones = context.selected_pose_bones if context.mode == "POSE" else context.selected_editable_bones
         selected_armatures = [sel_ob for sel_ob in context.selected_objects if sel_ob.type == "ARMATURE"]
--- a/addons/io_hubs_addon/components/operators.py
+++ b/addons/io_hubs_addon/components/operators.py
@@ -445,6 +445,14 @@ class CopyHubsComponent(Operator):
     panel_type: StringProperty(name="panel_type")
     component_name: StringProperty(name="component_name")
 
+    @classmethod
+    def poll(cls, context):
+        panel = getattr(context, 'panel')
+        panel_type = PanelType(panel.bl_context)
+        return panel_type == PanelType.OBJECT and context.mode == "OBJECT" \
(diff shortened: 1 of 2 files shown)


Exairnous:
Hey, nice!  I'd forgotten that this was present in the last version.
I found a couple issues and have some thoughts though.
- Copying won't work for bones that are present in the same armature.
- The operator is missing the {'REGISTER', 'UNDO'} options.
- Is there a reason why they're being limited to a single mode?  I think it should support Object, Edit, and Pose at least, and generally it's better UI practice to grey out/disable the button rather than hide it (hiding it on scenes seems good to me for the moment, but it might be nice to add a more traditional copy/paste for them in the future).
- Most Blender collection properties/lists support both keyword and index access, so when appending the bones to selected objects you can just do `selected_objects.append(target_armature_bones[target_bone.name])`
- Speaking of using a selected objects variable to store bones, I really think we should standardize on "host" as what we call the thing that has a component on it.  I've tried to do this in all of my code and I think it makes things much clearer because objects are a specific thing in Blender and sometimes passed along with what the component is on, plus I think it's much easier to remember that you may have to deal with multiple data types when the variable has "host" in it as opposed to "obj" or "ob" which suggest that you're dealing only with objects and not with scenes, materials, etc..  So I'd personally rename selected_objects/src_obj to selected_hosts/src_host.

keianhzo:
Thanks for the feedback. I think I have address all of it.

addons/io_hubs_addon/components/operators.py
```diff
@@ -437,6 +437,63 @@ def split_and_prefix_report_messages(report_string):
...
+            if host.type == "ARMATURE" and context.mode != "OBJECT":
+                selected_hosts.append(*self.get_selected_bones(context))
+            else:
+                selected_hosts.append(host)
+
+        return selected_hosts
+
+    def execute(self, context):
+        src_host = None
+        selected_hosts = self.get_selected_hosts(context)
```
Exairnous, line 472: I think this can just be initialized to a blank list?

addons/io_hubs_addon/components/ui.py
```diff
@@ -50,6 +50,15 @@ def draw_component(panel, context, obj, row, component_item):
 
         top_row.label(text=display_name)
 
+        if has_properties:
```
Exairnous, line 53: I think scene components can be skipped here since there's no current support for them?

Exairnous (commented):
Thanks.  I left a couple small comments, but overall it looks good.

addons/io_hubs_addon/components/operators.py
keianhzo: Yeah we can do it

addons/io_hubs_addon/components/ui.py
keianhzo: Actually I'm going to poll on the operator to disable in the unsupported cases better

keianhzo:
Updated. Thanks!

PR #198, opened 2023-02-15
Add a dynamic option for audio zones to be movable

keianhzo:
*Note*: To land after the new loader is active in the client.

This PR adds support for dynamic audio-zones.


Exairnous:
Just a small note on the checkbox tooltip/description.  "Wether" should be "Whether".

PR #200, opened 2023-02-15
Text-color-min-max

keianhzo:
Add missing color min/max text component values


Exairnous:
Did you mean to include the dynamic audio zone commit with this?  Other than that, it looks good to me.

PR #204, opened 2023-02-27
Migrate Y/Z axis swap and bump version

keianhzo:
Bump particle-emitter component version to 1.1.0 and add a Y/Z axis migration to be consistent with the new swapping

### DIFF
--- a/addons/io_hubs_addon/components/definitions/particle_emitter.py
+++ b/addons/io_hubs_addon/components/definitions/particle_emitter.py
@@ -1,9 +1,12 @@
 from bpy.props import FloatProperty, EnumProperty, FloatVectorProperty, StringProperty, IntProperty
 from ..hubs_component import HubsComponent
-from ..types import Category, NodeType, PanelType
+from ..types import Category, NodeType, PanelType, MigrationType
 from ..consts import INTERPOLATION_MODES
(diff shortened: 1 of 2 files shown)


addons/io_hubs_addon/components/definitions/particle_emitter.py
```diff
@@ -117,6 +120,25 @@ def gather(self, export_settings, object):
...
+            self.startVelocity = startVelocity
+
+            endVelocity = self.endVelocity.copy()
+            endVelocity = Vector((endVelocity.x, endVelocity.z, endVelocity.y))
+            self.endVelocity = endVelocity
+
+            if migration_type != MigrationType.GLOBAL or is_linked(ob) or type(ob) == bpy.types.Armature:
+                host_reference = get_host_reference_message(panel_type, host, ob=ob)
+                migration_report.append(
+                    f"Warning: The Particle Emitter component's Y and Z on startVelocity and endVelocity on the {panel_type.value} {host_reference} may not have migrated correctly")
```
Exairnous, line 138: I don't think a special warning needs to be displayed for linking/appending since this is just a normal version bump, although it's possible it may be needed on armatures that don't have an associated object, but I don't know.

PR #207, opened 2023-03-15
Add Alpha Cutoff property on Image component

korinVR:
Added the missing Alpha Cutoff property, which is necessary for the mask mode to function properly. Without this property, all pixels become opaque.

Related line on Hubs is here: https://github.com/mozilla/hubs/blob/7fb2d12fa2f6be1572c695de4c73792397b175c8/src/components/media-image.js#L147

Blender:
image

On Spoke:
image

### DIFF
--- a/addons/io_hubs_addon/components/definitions/image.py
+++ b/addons/io_hubs_addon/components/definitions/image.py
@@ -32,7 +32,7 @@ class Image(HubsComponent):
         description="Transparency Mode",
         items=TRANSPARENCY_MODE,
         default="opaque")
-    
+
     alphaCutoff: FloatProperty(
         name="Alpha Cutoff",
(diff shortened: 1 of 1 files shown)


matthewbcool:
Thanks so much for this PR korinVR!

keianhzo:
Thanks @korinVR! The test has not been updating accordingly but we want to cut a release so I'm going to land this and fix it in a follo-up PR. For the future there is a guide here about how to run and update tests: https://github.com/MozillaReality/hubs-blender-exporter/wiki/Environment-setup#testing

keianhzo (approved):
Thanks @korinVR! The test has not been updating accordingly but we want to cut a release so I'm going to land this and fix it in a follo-up PR. For the future there is a guide here about how to run and update tests: https://github.com/MozillaReality/hubs-blender-exporter/wiki/Environment-setup#testing

PR #212, opened 2023-03-28
Add support for default actions in loop-animation

keianhzo:
Closes https://github.com/MozillaReality/hubs-blender-exporter/issues/210

### DIFF
--- a/addons/io_hubs_addon/components/definitions/loop_animation.py
+++ b/addons/io_hubs_addon/components/definitions/loop_animation.py
@@ -837,28 +837,64 @@ class LoopAnimation(HubsComponent):
             migration_warning = False
             tracks = self.clip.split(",")
             for track_name in tracks:
+                nla_track = None
+                action = None
+
+                # check regular
                 try:
                     nla_track = ob.animation_data.nla_tracks[track_name]
                     track_type = "object"
                 except (AttributeError, KeyError):
                     try:
+                        action = ob.animation_data.action
+                        if not action_has_nla_track(ob, action) and track_name == action.name:
+                            track_type = "object"
+                        else:
+                            raise KeyError
+                    except (AttributeError, KeyError):
+                        action = None
+
+                if not (nla_track or action):
+                    # check shapekey
+                    try:
                         nla_track = ob.data.shape_keys.animation_data.nla_tracks[track_name]
                         track_type = "shape_key"
                     except (AttributeError, KeyError):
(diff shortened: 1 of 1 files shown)


Exairnous (changes requested):
This looks generally pretty good to me.  I've left some inline comments and then there are a couple of things I'll mention here:

- The comments in `class TrackPropertyType(PropertyGroup):` need updating as they are no longer correct with the new behavior.
- I think since the glTF add-on prioritizes NLA tracks over actions that actions should be added after NLA tracks to the `UpdateTrackContextMenu` and `TracksContextMenu` classes.
    - Also, because of this prioritization, I believe an `action_has_nla_track` function will need to be added to prevent actions being added to the component when they are already present as NLA tracks (as the glTF exporter will export the track instead of the action), e.g. `if menu_id not in menu_tracks and not action_has_nla_track(ob, action) and not has_action(...)`
```python
def action_has_nla_track(ob, action):
    nla_tracks_with_action = []
    for nla_track in ob.animation_data.nla_tracks:
        track_action_name = get_action_name(nla_track)
        if track_action_name == action.name:
            nla_tracks_with_action.append(nla_track.name)
    if hasattr(ob.data, 'shape_keys') and ob.data.shape_keys and ob.data.shape_keys.animation_data:
        for nla_track in ob.data.shape_keys.animation_data.nla_tracks:
            track_action_name = get_action_name(nla_track)
            if track_action_name == action.name:
                nla_tracks_with_action.append(nla_track.name)
    return any(nla_tracks_with_action)
```
- I think the `migrate` function should be updated so that actions will be migrated correctly from files before version 1.0.0

Also, regarding what I brought up at the add-on meeting:
- I've had a closer look at the `get_animation_name` function and I think it'll be fine.  It should allow passing through animation names that are present on child objects, so old files that were set up that way should work again, and the validation shouldn't cause any problems and may actually make any errors clearer in the exported file.
- And as to having an action specific icon, Christian and I were talking after the meeting about potentially scanning child objects for animations, so in that case I think it's better to maintain the information on whether it's a shape key or not.

keianhzo:
Thanks, I've addressed the nits and I think I'll land it and we can address the rest of the feedback in a follow-up PR.

Exairnous (approved):
As discussed at the last couple of add-on meetings, I've pushed commits to address the rest of my feedback.  I think everything should be good, so as long as you don't find anything, I think this is good to be merged :)

PR #213, opened 2023-03-29
Automatic navmesh generation support using recast

keianhzo:
Automatic navigation mesh generation support using recast. Based on the work from https://github.com/przemir/RecastBlenderAddon


addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,548 @@
...
+                'Failed to load shared library: %s\nPath to shared library: %s\n\nTraceback: %s' %
+                (str(e),
+                 libpathr, tracebackStr))
+            os.chdir(prevWorkingDir)
+            return {'FINISHED'}
+
+        os.chdir(prevWorkingDir)
+        print("nverts %i" % nverts)
+        print("ntris %i" % ntris)
+        print("cell size %f" % recastData.cellsize)
```
Exairnous, line 319: I think stuff like this isn't usually printed to the terminal, so should these print statements should be removed?

Exairnous (changes requested):
This is going to be so nice!  A couple things though.

- A visible component with visibility unchecked should be added along with the navmesh component.
- The max slope property should display degrees instead of radians (`subtype='ANGLE'` in the float property).
- I think what's in Spoke should be matched as closely as possible since there isn't any documentation and people are already familiar with what's in Spoke:
    - I think the only significant differences in property names/order are that the max step height is named max climb, and that max climb and max slope are swapped.
    - Some of the default values also don't match what's in Spoke, so I'll list the Spoke defaults here (defaults taken from: https://github.com/mozilla/Spoke/blob/master/src/editor/recast/recast.worker.js):
        - cell size: 0.166
        - cell height: 0.10
        - agent height: 1.70
        - agent radius: 0.5
        - maximum step: height 0.3
        - maximum slope: 45 degrees (0.785398 radians)
        - minimum region area: 1.00m2 (adding `unit='AREA'` in the float property makes it display m2 after the number)
        - region merge size: 20
        - edge max length: 12
        - edge max error: 1
        - verts per poly: 3
        - detail sample distance: 13
        - detail max sample error: 1
    - I think it would also be a good idea to hide the more advanced options by default, either by a checkbox or subpanel, so as not to confuse people.  The hidden options would be:
        - region merge size
        - edge max length
        - edge max error
        - verts per poly
        - detail sample distance
        - detail max sample error
    - It would also be nice to offer the auto cell size option.  I think this should work okay (patterned after the code in https://github.com/mozilla/Spoke/blob/master/src/editor/nodes/FloorPlanNode.js):
```python
def get_auto_cell_size():
    bounding_boxes = []
    for obj in bpy.context.selected_objects:
        if obj.type == 'MESH':
            bbox = [obj.matrix_world @ Vector(point) for point in obj.bound_box]
            bounding_boxes.extend(bbox)

    bound_box_x_coords = []
    bound_box_y_coords = []
    for point in bounding_boxes:
        bound_box_x_coords.append(point.x)
        bound_box_y_coords.append(point.y)
        
    size_x = abs(min(bound_box_x_coords) - max(bound_box_x_coords))
    size_y = abs(min(bound_box_y_coords) - max(bound_box_y_coords))
    area = size_x * size_y
    
    return pow(area, 1 / 3) / 50
```

- It might also be nice to add a reset to defaults button (since you can't just delete a floor plan element and re-add it to get the defaults back)

Exairnous:
Got some feedback from people at Taco Tuesday Testing Day on this patch and it was requested that the navmesh not be selected after generation and that the original selection be maintained to allow for easier tweaking.

Exairnous:
I've been thinking more about adding the visible component as well as the navmesh component and it would make sense to add the visible component as a dependency of the navmesh component, except that when the visible component got added it would be added with it's default of visible set to true, while the navmesh should have visible set to false.  So this got me to thinking that it might be nice to be able to override the default property values of dependencies.  Obviously this doesn't need to be addressed in this PR, but do you think setting dependency properties is something we should add?

keianhzo:
We can maybe override the `HubsComponent` `init` method to initialize dependencies.

Exairnous:
Yes, as we talked about in the add-on meeting, we'll need to rearrange the order `init` is called, but then that's a good place for it.

Jack Saturn and I found a bug when testing one day, it appears to error out generating the navmesh when certain non-mesh objects are present in a scene, e.g. reflection probes.  Putting the non-mesh objects in a separate collection and disabling it will work around this bug, but it would be ideal if they can just be skipped.

keianhzo:
@Exairnous I think I've address most of the feedback if you want to review again before I merge

Exairnous:
Yes, will review :)

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+        col = layout.column()
+        col.row().prop(recastPropertyGroup, "agent_height", text="Height")
+        col.row().prop(recastPropertyGroup, "agent_radius", text="Radius")
+        col.row().prop(recastPropertyGroup, "slope_max", text="Maximum slope")
+        col.row().prop(recastPropertyGroup, "climb_max", text="Maximum step height")
+
+        layout.label(text="Region:")
+        col = layout.column()
+        col.row().prop(recastPropertyGroup, "region_min_size", text="Min region size")
+        col.row().prop(recastPropertyGroup, "partitioning", text="Partitioning")
```
Exairnous, line 593: I completely forgot about this in the first review, but I think the partitioning property should be included in advanced settings too.

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+import ctypes
+import ctypes.util
+from ctypes import c_int, c_float
+
+import bmesh
+
+
+CELL_SIZE_DEFAULT = 0.166
+CELL_HEIGHT_DEFAULT = 0.10
+SLOPE_MAX_DEFAULT = radians(60)
```
Exairnous, line 40: Spoke has the maximum slope defaulting to 45 degrees.

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+CLIMB_MAX_DEFAULT = 0.9
```
Exairnous, line 41: Spoke has the maximum step height defaulting to 0.3.

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+EDGE_MAX_LENGTH = 12.0
+EDGE_MAX_ERROR = 1.0
+REGION_MIN_SIZE = 1.0
+REGION_MERGE_SIZE = 20.0
+VERTS_PER_POLY_DEFAULT = 3
+SAMPLE_DIST_DEFAULT = 13.0
+SAMPLE_MAX_ERROR_DEFAULT = 1.0
+PARTITIONING_DEFAULT = 'WATERSHED'
+COLOR_DEFAULT = (0.0, 1.0, 0.0, 1.0)
+AUTO_CELL_DEFAULT = False
```
Exairnous, line 53: Spoke defaults to auto cell size enabled, which I think is probably prudent to keep the generation time down on large meshes.

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+    color: FloatVectorProperty(name="Color",
+                               description="Color",
+                               subtype='COLOR_GAMMA',
+                               default=COLOR_DEFAULT,
+                               size=4,
+                               min=0,
+                               max=1)
+    expanded: BoolProperty(name="expanded", default=True)
+
+    auto_cell: BoolProperty(name="Auto Cell", default=AUTO_CELL_DEFAULT)
```
Exairnous, line 557: Nit: I think this should be named "Auto cell size" to match Spoke.

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+
+CELL_SIZE_DEFAULT = 0.166
+CELL_HEIGHT_DEFAULT = 0.10
+SLOPE_MAX_DEFAULT = radians(60)
+CLIMB_MAX_DEFAULT = 0.9
+AGENT_HEIGHT_DEFAULT = 1.70
+AGENT_RADIUS_DEFAULT = 0.5
+EDGE_MAX_LENGTH = 12.0
+EDGE_MAX_ERROR = 1.0
+REGION_MIN_SIZE = 1.0
```
Exairnous, line 46: Turns out the defaults I was looking at weren't the right ones for this property and Spoke actually defaults to a minimum region size of 4.0 m2.  So sorry, my bad.

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+        if not recastPropertyGroup.auto_cell:
+            col.row().prop(recastPropertyGroup, "cell_size", text="Cell size")
+        col.row().prop(recastPropertyGroup, "cell_height", text="Cell height")
+
+        layout.label(text="Agent:")
+        col = layout.column()
+        col.row().prop(recastPropertyGroup, "agent_height", text="Height")
+        col.row().prop(recastPropertyGroup, "agent_radius", text="Radius")
+        col.row().prop(recastPropertyGroup, "slope_max", text="Maximum slope")
+        col.row().prop(recastPropertyGroup, "climb_max", text="Maximum step height")
```
Exairnous, line 588: maximum slope and maximum step height should be switched to match Spoke's order.

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+    bl_idname = "recast.build_navigation_mesh"
+    bl_label = "Build Navigation Mesh"
+    bl_description = "Build navigation mesh from the selected objects using recast."
+    bl_options = {'REGISTER', 'UNDO'}
+
+    def execute(self, context):
+        # bpy.ops.wm.call_menu(name="ADDITIVE_ANIMATION_insert_keyframe_menu")
+
+        selected_objects = context.selected_objects
+        if len(selected_objects) == 0:
```
Exairnous, line 345: I think this should take into account only mesh objects, rather than all selected objects e.g.
```suggestion
        if len([obj for obj in selected_objects if obj.type == 'MESH']) == 0:
```

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+            self.report({'WARNING'}, 'No meshes selected')
+            return {"CANCELLED"}
+
+        from ..components.definitions.nav_mesh import NavMesh
+        nav_mesh_id = NavMesh.get_name()
+        for ob in context.selected_objects:
+            if has_component(ob, nav_mesh_id):
+                self.report({'ERROR'}, 'A Navmesh cannot be part of the selection')
+                return {'FINISHED'}
```
Exairnous, line 354: I think this should return {'CANCELLED'} to avoid pushing an undo step.

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+        navMeshes = get_objects_with_component(nav_mesh_id)
+        if navMeshes:
+            navMesh = navMeshes[0]
+
+        addon_prefs = get_addon_pref(context)
+        libpath = os.path.abspath(addon_prefs.recast_lib_path)
+        libpathr = libpath.replace("\\", "/")
+        if not os.path.exists(libpathr):
+            self.report({'ERROR'}, 'File not exists: %s\n' % libpathr)
+            return {'FINISHED'}
```
Exairnous, line 366: I don't think any permanent changes have been made at this point, so I think this should return {'CANCELLED'} to avoid pushing an undo step.

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+            # print("dmeshv1 %f" % dmeshv1)
+
+            createMesh(context, dmesh, obj=navMesh)
+
+        # what was allocated in C/C++ should be also deallocated there
+        recast.freeNavMesh(pmesh, dmesh, reportMsg, nreportMsg)
+
+        bpy.ops.object.select_all(action='DESELECT')
+        for obj in selected_objects:
+            obj.select_set(True)
```
Exairnous, line 430: I think the active object should also be saved/restored as well as the selected objects?

addons/io_hubs_addon/third_party/recast.py
```diff
@@ -0,0 +1,636 @@
...
+        default=PARTITIONING_DEFAULT)
+
+    color: FloatVectorProperty(name="Color",
+                               description="Color",
+                               subtype='COLOR_GAMMA',
+                               default=COLOR_DEFAULT,
+                               size=4,
+                               min=0,
+                               max=1)
+    expanded: BoolProperty(name="expanded", default=True)
```
Exairnous, line 555: I would have the advanced settings panel default to closed (False), so it doesn't distract from the most commonly changed settings.

Exairnous (changes requested):
I found a few things which I've noted in inline comments, but this is looking really good.

Also, is there a reason you made your own subpanel for the advanced settings, rather than using a blender one via defining another panel class and adding `bl_parent_id = "SCENE_PT_blendcast"` and `bl_options = {'DEFAULT_CLOSED'}` (if you agree with having it closed by default)?  I think it would fit in better stylistically here.

keianhzo:
@Exairnous Thanks for all the feedback! Addressed.

Exairnous (approved):
I added a commit to fix a couple small things, but other than that I think everything looks good.

PR #214, opened 2023-03-30
Fix encode call for Blender 3.5

keianhzo:
`ExportImage.encode` has changed it's signature in the Blender 3.5 glTF exporter. This PR adds 3.5.0 testing support and fixes the encoding issue.

### DIFF
--- a/addons/io_hubs_addon/io/utils.py
+++ b/addons/io_hubs_addon/io/utils.py
@@ -31,10 +31,13 @@ class HubsExportImage(gltf2_blender_image.ExportImage):
             export_image.fill_image(image, dst_chan=chan, src_chan=chan)
         return export_image
 
-    def encode(self, mime_type: Optional[str]) -> Union[Tuple[bytes, bool], bytes]:
+    def encode(self, mime_type: Optional[str], export_settings) -> Union[Tuple[bytes, bool], bytes]:
         if mime_type == "image/vnd.radiance":
             return self.encode_from_image_hdr(self.blender_image())
-        return super().encode(mime_type)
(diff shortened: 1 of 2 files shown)


addons/io_hubs_addon/io/utils.py
```diff
@@ -31,10 +31,13 @@ def from_blender_image(image: bpy.types.Image):
             export_image.fill_image(image, dst_chan=chan, src_chan=chan)
         return export_image
 
-    def encode(self, mime_type: Optional[str]) -> Union[Tuple[bytes, bool], bytes]:
+    def encode(self, mime_type: Optional[str], export_settings) -> Union[Tuple[bytes, bool], bytes]:
         if mime_type == "image/vnd.radiance":
             return self.encode_from_image_hdr(self.blender_image())
-        return super().encode(mime_type)
+        if bpy.app.version < (3, 5, 0):
```
netpro2k, line 37: It looks like this changed in 3.4 https://github.com/KhronosGroup/glTF-Blender-IO/commit/9af5fadf3ec6387590ccf5acf4036d81f3907dea

keianhzo: That commit was part of the 3.5.0 milestone: https://github.com/KhronosGroup/glTF-Blender-IO/pull/1794
It wasn't there when 3.4.0 was tagged AFAIK.

Exairnous: I can confirm that this is only needed for 3.5

Exairnous (approved):
LGTM

hrithikwins (commented):
.

PR #215, opened 2023-03-30
Show a warning if a nav mesh has more than one material slots assigned

### LINKED ISSUE #206: Add a warning if a nesh with a nav-mesh component has multiple materials
**Description**
If a mesh with the `nav-mesh` component has multiple material this will make the mesh to load as two meshes in Hubs, one per material, and in hubs only the first found mesh with a `nav-mesh` component will be used. This will be unexpected for the user.

**To Reproduce**
- Add a mesh
- Add two materials assigned to different meshes
- Add a `nav-mesh` component to the mesh
- Export the glb
- Load in Hubs and see how only part of the original nav mesh is used.

**Expected behavior**
That's actually the expected behavior as Hubs only supports one nav mesh but we should warn he user

**Additional context**
We should add a warning in the component's panel to flag that the nav mesh won't export correctly due to having multiple materials assigned.

keianhzo:
Closes #206

Also adds support for copying dependencies properties


Exairnous (changes requested):
Aside from the one comment I left, the nav mesh changes look good to me.

For the copy component changes, it seems to me like they could have been their own PR, but I think they should at least be mentioned in the commit message.  Aside from that, I feel like there could be a cleaner way to implement this.  With the current changes, the networked component is getting a copy button, which I don't think is beneficial.  It kinda feels like what's wanted is for dep only components with no properties to not to get a copy button and everything else to get a copy button, so how about `if has_properties or not component_class.is_dep_only():`?

Also, to allow drawing for components that don't have properties, I think you could just check if `component_item.expanded` is True (which should always be the case for components with no properties, I think?).

While testing, it occurred to me that it would also be nice if dependency settings were copied over when copying a parent, but that would be for a followup PR.

Exairnous:
Oh, I also just realized that with allowing dep only components to be copied, they should be skipped if the host it's going to doesn't have a component that depends on it otherwise you'll end up with components on hosts that aren't needed and can't be removed.

keianhzo:
@Exairnous I've addressed all the feedback and added support to copy deps properties. Ideally as you said this should have been a different PR but at this point I rather just get it all merged at once.

Exairnous (approved):
I pushed a commit to deal with that last issue I brought up, but otherwise I think it's good.  If you agree then go ahead and merge it :)

PR #216, opened 2023-03-30
Refactor OpenImage operator and reuse in environment-settings

keianhzo:
This PR adds a reusable `OpenImage` operator and adds it to the `environment-settings` components images. It also refactors the existing `OpenReflectionProbeEnvMap` to use `OpenImage`.

Closes https://github.com/MozillaReality/hubs-blender-exporter/issues/208


addons/io_hubs_addon/components/operators.py
```diff
@@ -500,6 +502,73 @@ def execute(self, context):
...
+    files: CollectionProperty(type=PropertyGroup)
+    filter_folder: BoolProperty(default=True, options={"HIDDEN"})
+    filter_image: BoolProperty(default=True, options={"HIDDEN"})
+    target_property: StringProperty()
+
+    relative_path: BoolProperty(
+        name="Relative Path", description="Select the file relative to the blend file", default=True)
+
+    disabled_message = "Can't open/assign images to linked data blocks. Please make it local first"
+    component = None
```
Exairnous, line 520: Unused?

Exairnous (changes requested):
So glad this is finally getting done!  Thanks :)
It mostly looks good, but there are a couple small things I've pointed out in inline comments.

I almost missed that the poll/description wasn't targeting the component host in the generic operator, which got me to thinking that we should probably try to find a way to make that more foolproof, either by adding in generic `get_host`/`get_component` functions that look for context pointers or we get in the habit of subclassing operators for each panel, or something like that.  Anyway, adding in generic host stuff doesn't need to happen for this PR, but it's something I think we should probably keep in mind/look into.

addons/io_hubs_addon/components/utils.py
```diff
@@ -131,7 +131,16 @@ def is_linked(datablock):
...
+def update_image_editors(old_img, img):
+    for window in bpy.context.window_manager.windows:
+        for area in window.screen.areas:
+            if area.type == 'IMAGE_EDITOR':
+                if area.spaces.active.image == old_img:
+                    area.spaces.active.image = img
+
 # Note: Set up stuff specifically for C FILE pointers so that they aren't truncated to 32 bits on 64 bit systems.
+
+
```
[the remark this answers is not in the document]
keianhzo: This is a PEP8 thing, it seems that it like two spaces there

addons/io_hubs_addon/components/definitions/environment_settings.py
```diff
@@ -87,6 +87,9 @@ def draw(self, context, layout, panel):
         row = layout.row(align=True)
         sub_row = row.row(align=True)
         sub_row.prop(data=self, property="backgroundTexture")
+        sub_row.context_pointer_set("hubs_component", self)
+        op = sub_row.operator("image.hubs_open_image", text='', icon='FILE_FOLDER')
+        op.target_property = "backgroundTexture"
```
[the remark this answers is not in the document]
keianhzo: Good catch, done

addons/io_hubs_addon/components/operators.py
```diff
@@ -500,6 +502,73 @@ def execute(self, context):
...
+
+    @ classmethod
+    def description(cls, context, properties):
+        description_text = "Load an external image "
+        if bpy.app.version < (3, 0, 0) and is_linked(context.active_object):
```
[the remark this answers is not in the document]
keianhzo: Good catch!

keianhzo:
@Exairnous I think I've addressed all feedback, let me know if you want to take another look, otherwise I will merge

addons/io_hubs_addon/components/utils.py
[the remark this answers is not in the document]
Exairnous: Ah, okay.  That's annoying, but if it's needed, it's needed.

Exairnous:
I'll take a quick look, but I expect it's good :)

addons/io_hubs_addon/components/definitions/environment_settings.py
```diff
@@ -95,9 +95,15 @@ def draw(self, context, layout, panel):
             sub_row.enabled = False
             add_link_indicator(sub_row, self.backgroundTexture)
 
+        sub_row.context_pointer_set("hubs_component", self)
+        sub_row.context_pointer_set("host", self.backgroundTexture)
+        op = sub_row.operator("image.hubs_open_image", text='', icon='FILE_FOLDER')
```
Exairnous, line 100: I missed this in the first review, but these should be on the `row` layout, not the `sub_row` layout.
Also, the host for this should be `context.scene`, not the image (we only need to disable stuff if the host and its component are linked)

addons/io_hubs_addon/components/definitions/environment_settings.py
```diff
@@ -106,6 +112,11 @@ def draw(self, context, layout, panel):
             sub_row.enabled = False
             add_link_indicator(sub_row, self.envMapTexture)
 
+        sub_row.context_pointer_set("hubs_component", self)
+        sub_row.context_pointer_set("host", self.envMapTexture)
+        op = sub_row.operator("image.hubs_open_image", text='', icon='FILE_FOLDER')
```
Exairnous, line 117: Same as above.

addons/io_hubs_addon/components/definitions/reflection_probe.py
```diff
@@ -759,8 +700,10 @@ def draw(self, context, layout, panel):
             sub_row = row.row(align=True)
             sub_row.enabled = False
             add_link_indicator(sub_row, self.envMapTexture)
-        row.operator("image.hubs_open_reflection_probe_envmap",
-                     text='', icon='FILE_FOLDER')
+        row.context_pointer_set("hubs_component", self)
+        row.context_pointer_set("host", self.envMapTexture)
```
Exairnous, line 704: The host for this should be `context.active_object` and not the image for the reasons stated above.

addons/io_hubs_addon/components/operators.py
```diff
@@ -500,6 +502,74 @@ def execute(self, context):
...
+    relative_path: BoolProperty(
+        name="Relative Path", description="Select the file relative to the blend file", default=True)
+
+    disabled_message = "Can't open/assign images to linked data blocks. Please make it local first"
+
+    @ classmethod
+    def description(cls, context, properties):
+        description_text = "Load an external image "
+        ob = getattr(context, 'host')
+        if bpy.app.version < (3, 0, 0) and is_linked(ob):
```
Exairnous, line 525: Annoyingly, the host attribute isn't getting passed to descriptions for me in Blender versions below 3.5, but it isn't really needed for versions 3.0 and up, so it'll mostly work if we utilize the if statement's short circuiting on 3.0+ (sadly, it looks like 2.93 won't work with the context pointer and so won't get the description).
`getattr` also isn't needed here, and I prefer not to use the variable name `ob` for hosts because it adds misleading preconceptions of what type the host is (I know it's a small thing, but I find it's really easy to forget that hosts can be things other than objects).
```suggestion
        if bpy.app.version < (3, 0, 0) and is_linked(context.host):
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -500,6 +502,74 @@ def execute(self, context):
...
+            description_text += f"\nDisabled: {cls.disabled_message}"
+
+        return description_text
+
+    @ classmethod
+    def poll(cls, context):
+        ob = getattr(context, 'host')
+        if is_linked(ob):
```
Exairnous, line 533: Pretty much the same as above, except the poll only fails for linked hosts in 2.93 (everything above that is fine), so the behavior ends up being correct, it just prints an annoying error message in the terminal when hovering over the button.
```suggestion
        if is_linked(context.host):
```

Exairnous (changes requested):
I found a couple things that I've left inline comments for.

We seem to keep running into issues with `context_pointer_set`, so I'm beginning to wonder if we should see if we can avoid using it by possibly subclassing more stuff.  On the other hand, it seems like it's gradually getting fixed, so I suppose we could just wait it out.

keianhzo:
@Exairnous I've added a fix for the context pointer issues in my last commit, I think that can help avoid the console errors

PR #220, opened 2023-04-20
Improve image component alpha cutoff

Exairnous:
Limits the minimum and maximum values to between 0-1.
Hides the alpha cutoff property in the UI when the transparency mode isn't set to binary/mask.

Note: The alpha cutoff property still exports regardless of the transparency mode, but would it be better not to export it when it isn't used?


keianhzo:
That would be better but it's ok for now.

keianhzo (approved):
LGTM

PR #224, opened 2023-06-05
Add visible component as a nav-mesh dependency

keianhzo:
Add the visible component as a dependency for nav-mesh and move init so it happens after the deps have been resolved


hrithikwins (approved):
reviewed the changes

Exairnous (approved):
I added the isDependency property to the migration, but otherwise this looks good to me.  If everything looks good to you, go ahead and merge it :)

PR #227, opened 2023-06-29
Add missing pdf support

keianhzo:
This PRs adds the missing pdf component


Exairnous (changes requested):
This doesn't seem to work with the current loader in Hubs, only the new loader.  To handle both, I think it's probably best to just add in a gather method that exports this as an image component for now, then when the current loader is removed the gather method can be as well.  A "Controls" checkbox is also needed for feature parity (it looks like this isn't yet supported in the new loader, but it doesn't cause any problems either).

It would also be nice if this had at least a simple plane gizmo with an A4 or US-Letter aspect ratio (portrait) to give the user some indication of the sizing/positioning.

keianhzo:
Oh yeah, me are actually not exposing the pdf component for the old loader... I think I'll just hold this until the new loader is released then.

Exairnous (commented):
The new loader likely won't be the default for a long time though, since I believe the plan is to first have it as an individual preference, and the ability to add PDFs is currently very hidden, so it would be really nice to make it more obvious with a dedicated component.

I've left an inline comment with most of the code that should address the feedback I had (I went with A4 for the gizmo aspect ratio), but to make it work you'd also need to change line 205 of `io/gltf_exporter.py` to `component_data[component_class.gather_name()] = data` and add 
```
@classmethod
def gather_name(cls):
    return cls.get_name()
```
to `hubs_component.py`

I needed to allow for the component name to be overridden on gather, so it required a few more changes than I thought it would, but I feel like we should allow overriding the component name anyway because it feels strange not to be able to override that when we can override the contents.

keianhzo:
@Exairnous Thanks for the feedback. I've added it to the PR. Reg the `gather_name` method override, why not just overriding `get_name`?

Exairnous (approved):
LGTM.  One potential nit, I was a bit curious about the dash to title changes and when investigating them I noticed that `__name__` didn't have `cls` in front of it in `def get_display_name(cls, default=__name__):`.

keianhzo:
Good catch, it was referencing the module `__name__`. It didn't really affect much as we always set the display name but I've fixed it.

PR #228, opened 2023-07-07
Fix for gtlf exporter python library path changes

rawnsley:
Addresses https://github.com/MozillaReality/hubs-blender-exporter/issues/225 in Blender 3.6 by fixing some of the import paths. I'm not sure of the implications for previous versions of Blender though or if there are any other compatibility issues, so I will keep this in draft for now.

### DIFF
--- a/addons/io_hubs_addon/io/utils.py
+++ b/addons/io_hubs_addon/io/utils.py
@@ -1,14 +1,14 @@
 import os
 import bpy
 from io_scene_gltf2.blender.com import gltf2_blender_extras
-from io_scene_gltf2.blender.exp import gltf2_blender_gather_materials, gltf2_blender_gather_nodes, gltf2_blender_gather_joints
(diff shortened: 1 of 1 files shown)


Exairnous (approved):
LGTM.

I did some checking on that format variable we were talking about at the add-on meeting and I found the commit + PR where it was removed.  It looks like that variable always evaluated to `'gltf_format'` so if you wanted to just use that string I think it would be fine, but I'll leave it up to you :)

Commit + PR where it was removed:
https://github.com/KhronosGroup/glTF-Blender-IO/commit/e63bb38e87832f18c22ff8cb76e779f179a32526
https://github.com/KhronosGroup/glTF-Blender-IO/pull/1832

rawnsley:
@Exairnous I took your advice and simplified that change.

keianhzo:
Thanks for this fix!

Exairnous:
@rawnsley Ah, nice! :slightly_smiling_face:

PR #231, opened 2023-07-12
Add component to class registry during register_component

keianhzo:
Add class to the registry inside register_component so external add-ons using the registry can use only one function call.


keianhzo:
Better leave this for a later time when we consolidate a public facing API for the registry

PR #232, opened 2023-07-16
Revert some poll changes in commit "Disable operators for linked hosts fixes."

Exairnous:
Revert the changes to the AddHubsComponent poll and the poll default return values from commit fa0f4761b8421b1b20336634704b4223b334bda9

The changes to the AddHubsComponent poll caused an intermittent bug when adding components where the component wouldn't be added when selected because of a poll failure. Things like adding an object and then trying to add a component would trigger this bug. For some reason, adding an object seems to cause problems with passing the context pointer.

The changes to the poll default return values were because I thought it made more sense to return false when the poll was in an invalid state because the context pointer wasn't set, but given the above bug, reverting to returning true seems safer.


keianhzo (approved):
LGTM, Thanks!

PR #242, opened 2023-09-14
Improve the file loading guard on the undo stack handler.

### LINKED ISSUE #239: Blender won't start after addon installation
**Description**
I installed the Hubs Blender addon 1.2.0.190 with Blender 3.6.2. It works fine when i start blender, install it and then work in the session. But when i close Blender and start it again, it won't start as long as the addon is installed. Blender starts loading but then it closes after half a sec without an error message. Even when i deactivate the addon. Only option to start blender is with the "blender_factory_startup".  When i start blender with the blender_factory_startup and uninstall the Hubs Blender addon, blender starts afterwards without problems.   

**To Reproduce**
Install the addon with Blender 3.6.X and restart Blender. 
But i'm sure it is not that simple.. There must be an issue with my system.
I'm not sure but the problem started with the latest blender update from 17.08. Blender 3.6.2 **LTS** Maintainance Release.

**Expected behavior**
Blender starts with the addon installed.

**Screenshots**

**Hardware** 
 - Add-on Release Version: Hubs Blender addon 1.2.0.190
 - Blender Version: Blender 3.6.2
I tried the steam version and the download version (here i tried 3.3.10, 3.6.0, 3.6.1, 3.6.2).
 I use Win11 (22H2).

**Additional context**
*Scene Links, models, any other helpful context*

Exairnous:
Use a more flexible approach so that the guard will be certain to remain in place until the blend file has loaded to a usable state. This fixes #239.

### DIFF
--- a/addons/io_hubs_addon/components/handlers.py
+++ b/addons/io_hubs_addon/components/handlers.py
@@ -239,11 +239,11 @@ def undo_stack_handler(dummy, depsgraph):
     global object_data_switched
 
     # Return if Blender isn't in a fully loaded state. (Prevents Blender crashing)
-    if file_loading and not bpy.context.space_data:
-        file_loading = False
-        return
+    if file_loading:
+        if not bpy.context.space_data:
+            return
 
-    file_loading = False
+        file_loading = False
 
     # Get a representation of the undo stack.
     binary_stream = io.BytesIO()


keianhzo (approved):
LGTM

PR #247, opened 2023-11-07
Don't fail if a dependency does not exist

keianhzo:
If a dependency is already removed we are failing to remove and a confusing  stacktrace is shown to the user. This PR makes us continue if a dependency doesn't exist maybe because it' been already manually removed.

### DIFF
--- a/addons/io_hubs_addon/components/utils.py
+++ b/addons/io_hubs_addon/components/utils.py
@@ -30,7 +30,7 @@ def add_component(obj, component_name):
                 if not dep_exists:
                     add_component(obj, dep_name)
             else:
-                print("Dependency '%s' from module '%s' not registered" %
+                print("Dependency '%s' from component '%s' not registered" %
                       (dep_name, component_name))
         component_class.init(obj)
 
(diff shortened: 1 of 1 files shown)


Exairnous:
But why is the dependency not there in the first place?  There should be no way for a user to remove a dependency through the UI.  Is this related to components in the BG add-on like `BG Grabbable` with its `BG RigidBody` and `BG Physics Shape` dependencies?  If so and these are actual hard dependencies, then I think they need to have the `isDependecy` property set in their `init` function like so (or perhaps this property should be automatically set?):
```
    @classmethod
    def init(cls, obj):
        obj.hubs_component_list.items.get('rigidbody').isDependency = True
```
Otherwise, if they aren't hard dependencies, then we should probably introduce the concept of soft dependencies (e.g. components that are added with the main component that are likely to be useful, but aren't strictly required) and switch them to be soft.

keianhzo:
Oh yeah, i forgot that we had that flag. That will make it. I'll close this. Thanks

Exairnous:
You're welcome!  Happy to help :)

PR #248, opened 2023-11-17
Alignment and scale options for media frames

rawnsley:
The addition of a media frame option to disable the default scale-to-fit-bounds behaviour and options to control how media should be aligned within the frame when it doesn't fill the space exactly. For example, these settings would place objects on the floor of the media frame and allow to keep their original scale:

Screenshot 2023-11-16 at 14 22 15

This companion hubs-client PR contains some concrete examples.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/media_frame.py
+++ b/addons/io_hubs_addon/components/definitions/media_frame.py
@@ -59,13 +59,6 @@ class MediaFrame(HubsComponent):
         'version': (1, 0, 0)
     }
 
-    bounds: FloatVectorProperty(
-        name="Bounds",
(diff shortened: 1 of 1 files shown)


Exairnous:
Thanks for this @rawnsley !
I was hoping to test this out over the weekend, but didn't end up getting to it.  I plan to test it in the next day, but it looks alright just glancing over it.  The one thing that occurs to me is that I think the two `if export_settings['gltf_yup']:` sections could be combined.

keianhzo (approved):
LGTM. Thanks for the contribution!

keianhzo (approved):
LGTM. Thanks for the contribution!

PR #252, opened 2023-12-01
Tests updates

keianhzo:
Tests are failing for me on 3.6.0, it seems that the culprit is the way we enable the add-on for instrumentation Blender. I've replaced that with the  BLENDER_USER_SCRIPTS command line argument that seems to work more reliably across versions.

So this fixed the test locally but somehow they still fail here, not sure why they started to suddenly fail though...


keianhzo:
It turned out not to be related to this. Closing

PR #253, opened 2023-12-11
Scene Debugger followup PR

Exairnous:
Displays the forced export options to the user in the UI along with the reasoning behind the hard-coded values in their tooltips and adds support for Blender versions below 3.2.

Depends on https://github.com/MozillaReality/hubs-blender-exporter/pull/249

Before:

After:

### DIFF
--- a/addons/io_hubs_addon/debugger.py
+++ b/addons/io_hubs_addon/debugger.py
@@ -77,8 +77,10 @@ def export_scene(context):
         'use_active_collection': export_prefs.use_active_collection,
         'export_apply': export_prefs.export_apply,
         'export_force_sampling': False,
-        'use_active_scene': True
     }
+    if bpy.app.version >= (3, 2, 0):
+        args['use_active_scene'] = True
+
     bpy.ops.export_scene.gltf(**args)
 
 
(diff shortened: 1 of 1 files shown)


keianhzo:
Hey @Exairnous can you rebase with the latest master and fix the conflicts? Thanks!

Exairnous:
Rebased.  Thanks!

addons/io_hubs_addon/debugger.py
```diff
@@ -77,8 +77,10 @@ def export_scene(context):
         'use_active_collection': export_prefs.use_active_collection,
         'export_apply': export_prefs.export_apply,
         'export_force_sampling': False,
-        'use_active_scene': True
     }
+    if bpy.app.version >= (3, 2, 0):
```
keianhzo, line 81: Just curious does this do any harm if < 3.2.0? Does it crash?

Exairnous: The `use_active_scene` export option isn't present in the glTF add-on in Blender versions under 3.2, so trying to include it causes the export to fail in those versions.

PR #257, opened 2023-12-14
Improve accessibility for the display of the active room's flags.

Exairnous:
This changes them from red and green dots to use an X to represent not active, a checkmark to represent active and a line to represent the room not being open.  The default Blender icons are used so they will take their colors from the current theme as well.


keianhzo (approved):
LGTM Thanks for this!

PR #258, opened 2023-12-14
Improve room flag descriptions and include the actual flag at the end.

Exairnous:
Improve the room flag descriptions to be more informative on what they do and add the actual query string parameter at the end so that it can be looked up in the documentation.

Feel free to edit the descriptions further if you see mistakes or think of improvements.


keianhzo (approved):
LGTM Thanks!

PR #260, opened 2023-12-15
Init on addon register

keianhzo:
We were not initializing on add-on register, only on scene open. This PR ensures that the scene debugger is always initialized after  addon registreation.


Exairnous (approved):
Code LGTM.  Not sure why the tests aren't passing here, but they're fine when I ran them locally.

keianhzo:
Not sure why but it seemed to be the operator call while registering, maybe the operator was not yet fully initialized.

PR #261, opened 2023-12-16
Avoiding confusing between Blender X axis and Threejs Y Axis

hrithikwins:
Fixes the issue mentioned in https://github.com/MozillaReality/hubs-blender-exporter/issues/217

### DIFF
--- a/addons/io_hubs_addon/components/definitions/billboard.py
+++ b/addons/io_hubs_addon/components/definitions/billboard.py
@@ -15,4 +15,4 @@ class Billboard(HubsComponent):
     }
 
     onlyY: BoolProperty(
-        name="Vertical Axis Only", description="Locks the Vertical Axis to enable only side to side movement", default=False)
+        name="Vertical Axis Only", description="Locks the Vertical Axis to enable only side to side movement in world space and removes any other rotational transforms", default=False)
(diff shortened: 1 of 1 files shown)


hrithikwins:
image

keianhzo (commented):
Not sure if "Only Horizontal" is preferred over the proposed "Vertical Axis Only". Any ideas @j-conrad @Exairnous ?

hrithikwins:
Thanks @keianhzo @Exairnous @j-conrad for pointing out the wrong commits, now this is fixed with 
image

Exairnous:
Thank you for the update @hrithikwins, this is looking much better!  We discussed in the add-on meeting that there should probably be a reference to it affecting the world space and removing the other rotational transforms, so what would you (and the others) think of adding "in world space and removes any other rotational transforms" to the end of the tooltip?

hrithikwins:
Sure, so that would mean that the tooltip contains the text as 

Locks the Vertical Axis to enable only side to side movement in world space and removes any other rotational transforms

Exairnous:
> Sure, so that would mean that the tooltip contains the text as
> 
> Locks the Vertical Axis to enable only side to side movement in world space and removes any other rotational transforms

Yes, that is what I'm suggesting :)

hrithikwins:
Thanks Exairnous , the latest commit has the more descriptive description :)

Exairnous:
Thank you Hrithik.

@keianhzo @j-conrad is everybody happy with the current wording?

hrithikwins:
@keianhzo @j-conrad

keianhzo (approved):
LGTM! Thank you @hrithikwins

PR #265, opened 2024-01-15
Scene Debugger: Allow spawning objects when not signed in.

Exairnous:
This restores the ability to spawn objects when not signed in and not using debugLocalScene that was accidentally removed in commit 83778e7b96d7b6ec6fda91cda8dd9dd37b93deb6


keianhzo (approved):
LGTM, Thank you!

PR #266, opened 2024-02-06
Fixed a bug when loading images using relative paths [Deprecated]

Spiderguy-F:
using the
class OpenImage(Operator)
fails when using the default 'relative_path' property (True).
We fixed that error and made the file browser start where the previous image was located. We also modified the dedfault location of the File Viewer to either:
a) Start where a file was selected the last time the File View was used or:
b) if an image had been previously assigned to the target_property, the File View will start with that image selected.
This is Blender's default behaviour

grafik
It probably failed when selecting an image for a reflection probe as well and should be working everywhere now.

### DIFF
--- a/addons/io_hubs_addon/components/operators.py
+++ b/addons/io_hubs_addon/components/operators.py
@@ -650,8 +650,9 @@ class OpenImage(Operator):
         old_img = self.hubs_component[self.target_property]
 
         # Load/Reload the first image and assign it to the target property, then load the rest of the images if they're not already loaded. This mimics Blender's default open files behavior.
+        filepath = os.path.join(self.directory, self.files[0].name)
         primary_img = bpy.data.images.load(
-            filepath=self.filepath, check_existing=True)
+            filepath=filepath, check_existing=True)
         primary_img.reload()
         self.hubs_component[self.target_property] = primary_img
 
--- a/addons/io_hubs_addon/components/operators.py
+++ b/addons/io_hubs_addon/components/operators.py
@@ -606,6 +606,7 @@ class OpenImage(Operator):
     bl_label = "Open Image"
     bl_options = {'REGISTER', 'UNDO'}
 
+    directory: StringProperty()
     filepath: StringProperty(subtype="FILE_PATH")
     files: CollectionProperty(type=PropertyGroup)
     filter_folder: BoolProperty(default=True, options={"HIDDEN"})
@@ -640,11 +641,7 @@ class OpenImage(Operator):
         layout.prop(self, "relative_path")
 
     def execute(self, context):
-        #dirname = os.path.dirname(self.filepath) #fails if path selected in the Blender File View is relative (starts with //)
-        abs_dir = bpy.path.abspath(self.filepath) # converts relative paths to absolute ones eg //MyPath -> C:\MyPath
-        #or:
-        #abs_dir = self.filepath if not self.relative_path else bpy.path.abspath(self.filepath)
-        dirname = os.path.dirname(abs_dir)
+        #dirname = os.path.dirname(self.filepath) #dirname fails if path selected in the Blender File View is relative (starts with //)
 
         if not self.files[0].name:
             self.report({'INFO'}, "Open image cancelled.  No image selected.")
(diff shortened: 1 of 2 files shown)


addons/io_hubs_addon/components/operators.py
```diff
@@ -640,7 +641,7 @@ def draw(self, context):
         layout.prop(self, "relative_path")
 
     def execute(self, context):
-        dirname = os.path.dirname(self.filepath)
+        #dirname = os.path.dirname(self.filepath) #dirname fails if path selected in the Blender File View is relative (starts with //)
```
Exairnous, line 644: I think this line can be completely removed.  Also, dirname seems to handle relative paths fine on my system (Linux), so maybe it's OS dependant.

addons/io_hubs_addon/components/operators.py
```diff
@@ -649,23 +650,23 @@ def execute(self, context):
         old_img = self.hubs_component[self.target_property]
 
         # Load/Reload the first image and assign it to the target property, then load the rest of the images if they're not already loaded. This mimics Blender's default open files behavior.
-        primary_filepath = os.path.join(dirname, self.files[0].name)
```
Exairnous, line 652: This line still needs to be here to match Blender's default behavior when opening multiple files, just replacing dirname with self.directory should work.

If you try out e.g. loading an image in an image texture node in a shader you should see that Blender always opens the first of the selected files, regardless of what self.filepath is.  If this isn't the case on your system then we probably need to account for OS differences.

addons/io_hubs_addon/components/operators.py
```diff
@@ -649,23 +650,23 @@ def execute(self, context):
...
         primary_img = bpy.data.images.load(
-            filepath=primary_filepath, check_existing=True)
+            filepath=self.filepath, check_existing=True)
```
Exairnous, line 654: To match Blender's default behavior this should use primary_filepath instead of self.filepath.

Exairnous (changes requested):
That's weird!  Relative paths work fine on my system (Linux), but apparently not everywhere so thanks very much for the fix!

This patch changes the behavior when opening multiple files, though (at least on my system).  Instead of opening the first file in the selected list, it opens the one specified by the filepath.  While I agree this makes more sense, I think it's more important to remain consistent in this case so as not to confuse users even more than Blender is already doing.

I'm going to mark this as changes requested for now, but feel free to provide a counter argument and we'll see what @keianhzo and @j-conrad think.

Also, thanks for the change to the default location of the file viewer.  That makes it much nicer.

addons/io_hubs_addon/components/operators.py
Spiderguy-F: I think in the 2. commit, I changed all that. It should have also restored the behaviour of the 1. file being selected, s.a.

addons/io_hubs_addon/components/operators.py
Exairnous: > n windows an absolute path always starts with :.

Ah, yes, that's probably it.

> I didn't change anything to the behavior of the import of the other lines but the filepath.

You did change the behavior of the import because `self.filepath` isn't necessarily going to be the same as `self.directory + self.files[0].name`.  Try adding an image texture and selecting five files from the bottom up, you'll see the bottom file in the browser field, but the top file will be the one that actually goes into the image texture (unless Blender thinks you're trying to import a sequence, in which case the behavior may be different).  Now, it's true that this isn't quite 100% consistent behavior within Blender (but it's close), and the way Blender does it isn't as intuitive (but I think you're pretty safe in assuming that the top selected file, is the one that gets loaded into the field when opening multiple files at a time), so let's see what @keianhzo and @j-conrad think, but if the behavior is changed then the comment needs updating as well to reflect the new behavior and that it differs from Blender's default.

Re: the `def draw` function.
I don't remember exactly why I had this there initially, but it was probably because it was easy to add and would prevent mistakes if anyone needed to add in messages or more complex UI elements (plus you also don't have to worry about whether the properties are hidden or not).

> @Exairnous Yes, the line can be removed, I uploaded a 2. commit where I removed it.

Weird, the line is still there for me, just commented out.  Don't get me wrong, the explanation is useful to know, but I feel like it'd be more useful in a best practices document, rather than on a commented out line of code in a source file.  Actually, this line is used in other places, so they'll need updating as well.

addons/io_hubs_addon/components/operators.py
Exairnous: The behavior is still changed, see my above comment.

addons/io_hubs_addon/components/operators.py
Spiderguy-F: In the 2. commit (Better solution for open image with relative path failing) I changed it back and removed that line.

addons/io_hubs_addon/components/operators.py
Spiderguy-F: Sorry, I'm new to pull requests. This commit: Better solution for open image with relative path failing had the original behaviour restored.

Spiderguy-F: Please consider this request obsolete, I created too much chaos with the branches and opened a new request for this topic, which includes your change requests
https://github.com/MozillaReality/hubs-blender-exporter/pull/266

keianhzo:
Closing in favor of https://github.com/MozillaReality/hubs-blender-exporter/pull/269

PR #269, opened 2024-02-15
Fix Open Image Relative Path failing - New Try

Spiderguy-F:
We cleaned up the chaos by creating a new branch. The behaviour should match the previous behaviour now, let's call it "intuitive". 
class OpenImage(Operator)
fails when using the default 'relative_path' property (True) on Windows. Python os.path.join on Windows does not recognize '//' to be the beginning of a directory and treats it like a filename.
We fixed that error and made the file browser start where the previous image was located. We also modified the dedfault location of the File Viewer to either:
a) Start where a file was selected the last time the File View was used or:
b) if an image had been previously assigned to the target_property, the File View will start with that image selected.
This is Blender's default behaviour
grafik

It probably failed when selecting an image for a reflection probe as well and should be working everywhere now.

### DIFF
--- a/addons/io_hubs_addon/components/operators.py
+++ b/addons/io_hubs_addon/components/operators.py
@@ -642,7 +642,6 @@ class OpenImage(Operator):
     
     def execute(self, context):
         #dirname = os.path.dirname(self.filepath) #dirname fails if path selected in the Blender File View is relative (starts with //) on Windows
-        print("filepath:", self.filepath)
         if not self.files[0].name:
             self.report({'INFO'}, "Open image cancelled. No image selected.")
             return {'CANCELLED'}
--- a/addons/io_hubs_addon/components/operators.py
+++ b/addons/io_hubs_addon/components/operators.py
@@ -606,11 +606,12 @@ class OpenImage(Operator):
     bl_label = "Open Image"
     bl_options = {'REGISTER', 'UNDO'}
 
+    directory: StringProperty()
     filepath: StringProperty(subtype="FILE_PATH")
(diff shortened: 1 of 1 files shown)


addons/io_hubs_addon/components/operators.py
```diff
@@ -634,38 +635,41 @@ def poll(cls, context):
                 return False
 
         return True
-
-    def draw(self, context):
-        layout = self.layout
+    
+    def draw(self, context): #this def is obsolete, bc. target_property is hidden by default now and relative_path is displayed by default anyways
```
keianhzo, line 639: Is this is not required anymore in any of the supported Blender versions, I think we can safely remove it.

addons/io_hubs_addon/components/operators.py
```diff
@@ -634,38 +635,41 @@ def poll(cls, context):
...
+        layout = self.layout 
         layout.prop(self, "relative_path")
-
+    
     def execute(self, context):
-        dirname = os.path.dirname(self.filepath)
-
+        #dirname = os.path.dirname(self.filepath) #dirname fails if path selected in the Blender File View is relative (starts with //) on Windows
```
keianhzo, line 644: This comment can be removed

keianhzo (changes requested):
Thanks for this! Looks good to me, just a couple of nits and also there are some formatting issues, please run `python ./check_style.py` and fix those to keep the code format consistency.

keianhzo:
Thanks perfect, thanks!

keianhzo (approved):

Issue #199, opened 2023-02-15
Make audio-params override to be disabled by default


keianhzo:
**Is your feature request related to a problem? Please describe.**
When you add any audio element, the audio-params component is added too. By default this component override property is enabled. It will make more sense to keep override disabled by default, this is also how Spoke behaves by default

**Describe the solution you'd like**
Make override property disabled by default. we will need to migrate < 1.0.0 versions and set it to False. For beta versions we can't really know if it was set to True in purpose so probably easiest to not make assumptions and respect the incoming property value.

Exairnous:
I have a couple thoughts/observations on this issue.
- I think defaulting the audio-params component to not override is fine, but I think the audio-target and audio-zone components should override that default when added and set it to override the audio because that is their main function.
- As far as I can tell, add-on versions before the override was introduced always overrode the audio, so they should have the override set to True.
- Since this was a new feature introduced near the end of the betas, it might be best to have the migration include versions less than or equal to 1.0.0 and just check whether that field is present, if it is then respect its value, if it isn't then set it to True.

keianhzo:
Yes, I think all that's reasonable.

PR #275, opened 2024-04-05
Switch the audio-params override to be disabled by default.

Exairnous:
Make audio zones and audio targets specifically enable the override because that's their whole purpose. When migrating, check to see if the field is present, if it is then its value is respected otherwise the override is set to true in order to preserve the previous state.

This addresses issue #199  It is also needed for properly importing the audio-params component in #63


keianhzo (approved):
LGTM

PR #281, opened 2024-04-22
Operator to automatically bake lightmaps for selected objects.

GottfriedHofmann:
Added an operator that bakes lightmaps using Cycles for all selected objects that have at least one material. It uses UV slot 1 for the lightmap and creates the slot if not present already. It also automatically creates the required MOZ_Lightmap nodes and wires them correctly. If a MOZ_Lightmap node is present, the addon assumes that the user has taken care of the correct setup and bakes directly.
Baking happens using the Cycles render engine. The created lightmaps are saved as .hdr to Blender's temporary directory, packed and then removed to keep the user's system clean.
The new operator can be found in the Object Properties Editor in a new Panel "Hubs Lightmap Baker"
image


Exairnous (changes requested):
Thank you for the PR and apologies for the delay in reviewing it.  I think it will provide some good functionality, but I think it's scope should be modified slightly.  I think your PR is great for baking light maps for objects that don't have them yet, but your PR doesn't address all the intricacies that can arise when attempting to rebake already set up light maps.  PR #48 does this very well and has a lot of error checking already in place, so I'm thinking it makes the more sense to have your PR to bake an initial light map for objects and let the other PR handle any rebaking.

With that in mind, I'm thinking that a check should be added to this PR to avoid baking light maps when they are already present on an object.

PR #48 also adds a panel in the render settings, so I'm thinking that maybe it would be better to add the operator there once that's in, rather than to a new panel in the object properties.

I believe calling bake via execute instead of invoke causes the UI to block with no feedback to the user.  I think it would be better to use the non-blocking version and reset all the settings after the bake has completed via a handler (like what is done with reflection probes).  The overridden settings can be stored and restored as with reflection probes in `reflection_probe.py` lines 362-379 and lines 334-337

`bpy.app.handlers.object_bake_complete` is probably the needed handler.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +670,174 @@ def invoke(self, context, event):
...
+        # Update the view layer so all parts take notice of the changed UV layout
+        bpy.context.view_layer.update()
+
+        # Gather all materials on the selected objects
+        materials = []
+        for obj in mesh_objs:
+            if len(obj.material_slots) >= 1:
+                # TODO: Make more efficient
+                for slot in obj.material_slots:
+                    if slot.material not in materials:
```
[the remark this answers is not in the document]
GottfriedHofmann: I do so in line 737: if len(obj.material_slots) >= 1:

If there is no material present it does not bake the object but removes it from selection.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +670,174 @@ def invoke(self, context, event):
...
+            # Remove non-mesh objects from selection to ensure baking will work
+            ob.select_set(False)
+
+        # set up UV layer structure. The first layer has to be UV0, the second one UV1 for the lightmap.
+        for obj in mesh_objs:
+            obj_uv_layers = obj.data.uv_layers
+            # Check whether there are any UV layers and if not, create the two that are required.
+            if len(obj_uv_layers) == 0:
+                obj_uv_layers.new(name='UV0')
+                obj_uv_layers.new(name='UV1')
```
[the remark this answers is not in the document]
GottfriedHofmann: I took the UV1 name from several tutorials I read. Since they all used UV1 as name for the slot I thought it would be save to do so as well. But "Lightmap" is a lot more descriptive so let's go with that.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +670,174 @@ def invoke(self, context, event):
...
+        context.scene.cycles.samples = self.samples
+        # Baking needs to happen without the color pass because we only want the direct and indirect light contributions
+        bake_settings = context.scene.render.bake
+        bake_settings.use_pass_direct = True
+        bake_settings.use_pass_indirect = True
+        bake_settings.use_pass_color = False
+        # The should be small because otherwise it could overwrite UV islands
+        bake_settings.margin = 2
+        # Not sure whether this has any influence
+        bake_settings.image_settings.file_format = 'HDR'
```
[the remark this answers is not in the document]
GottfriedHofmann: Forgot about that, will fix.

GottfriedHofmann:
I have created this addon with PR https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/48 in mind.
My addon currently uses the following logic:
Is a material present but no lightmap? - build the lightmap setup
Is a material with lightmap present? - assume the user has set up the light map correctly and re-bake.

addons/io_hubs_addon/components/operators.py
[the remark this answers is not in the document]
vincentfretin: The condition doesn't check if there are materials used. You may have one material slot but not assigned to material.
Why not just iterating over obj.data.materials here?
```
for obj in mesh_objs:
    for material in obj.data.materials:
        if material not in materials:
```
should work correctly.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,176 @@ def invoke(self, context, event):
...
+
+            # The layer for the lightmap needs to be the active one before lightmap packing
+            obj_uv_layers.active = obj_uv_layers[LIGHTMAP_LAYER_NAME]
+
+        # run UV lightmap packing on all selected objects
+        bpy.ops.object.mode_set(mode='EDIT')
+        bpy.ops.mesh.select_all(action='SELECT')
+        # TODO: We need to warn the user at some place like the README that the uv_layer[1] gets completely overwritten if it is called 'UV1'
+        # bpy.ops.uv.lightmap_pack()
+        bpy.ops.uv.smart_project()
```
vincentfretin, line 730: You create an image texture per material that is not shared between several materials if I'm not mistaken. Don't we need to do the smart_project for each group of objects sharing the same material instead?

I'm commenting on this PR because I'm doing my own Python script from the command line that I started before your PR that I'm continuing now, so I'm comparing with what I have.
I'm using 
`bpy.ops.uv.smart_project(island_margin=island_margin, correct_aspect=True)`
with island_margin=0.03 or 0.001 it depends, that fixed some black artifacts on some of my scene.
Changing bake_settings.margin like you do below may also fix the issue. I just wanted to share that.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,176 @@ def invoke(self, context, event):
...
+            node.image.filepath_raw = file_path
+            node.image.file_format = 'HDR'
+            node.image.save()
+            node.image.pack()
+            # Update the filepath so it unpacks nicely for the user.
+            # TODO: Mechanism taken from reflection_probe.py line 300-306, de-duplicate
+            new_filepath = f"//{node.image.name}.hdr"
+            node.image.packed_files[0].filepath = new_filepath
+            node.image.filepath_raw = new_filepath
+            node.image.filepath = new_filepath
```
vincentfretin, line 797: I tested those four lines in my own script on blender 4.2.1, it doesn't seem to be needed if you want to unpack the textures later, the names are correct in the textures folder as far as I can tell, and actually those lines gave me an error during the execution saying the file didn't exist. I didn't test your PR though, so may not apply, but that's something to verify.

addons/io_hubs_addon/components/operators.py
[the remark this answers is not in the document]
vincentfretin: Ah actually it may not work, obj.data.materials may give an error
AttributeError: 'PointLight' object has no attribute 'materials'
I had that condition in my code
`if obj.type == 'MESH' and getattr(ob.data, "materials", None)`
so yeah probably do what @Exairnous suggested:

```
for obj in mesh_objs:
    for slot in obj.material_slots:
        if slot.material is not None and slot.material not in materials:
            materials.append(slot.material)
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,176 @@ def invoke(self, context, event):
...
+    bl_label = "Bake Lightmaps"
+    bl_description = "Bake lightmaps of selected objects using the Cycles render engine and pack them into the .blend."
+    bl_options = {'REGISTER', 'UNDO'}
+
+    default_intensity: FloatProperty(name="Lightmaps Intensity",
+                                     default=3.14,
+                                     description="Multiplier for hubs on how to interpert the brightness of the image. Set this to 1.0 if you have set up the lightmaps manually and use a non-HDR format like png or jpg.")
+    resolution: IntProperty(name="Lightmaps Resolution",
+                            default=2048,
+                            description="The pixel resoltion of the resulting lightmap.")
```
vincentfretin, line 685: resoltion -> resolution

addons/io_hubs_addon/components/operators.py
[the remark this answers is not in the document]
GottfriedHofmann: Thank you, tested and it works! (thankfully I am already testing for mesh objects further up)

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +670,174 @@ def invoke(self, context, event):
...
+        return {'FINISHED'}
+
+    def invoke(self, context, event):
+        if bpy.data.is_saved is False:
+            def draw(self, context):
+                self.layout.label(
+                    text="You ned to save the .blend file before running this operator.")
+            bpy.context.window_manager.popup_menu(
+                draw, title="File not saved.", icon='ERROR')
+            return {'CANCELLED'}
```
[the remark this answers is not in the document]
GottfriedHofmann: Tested and it is not needed indeed 🙏

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,176 @@ def invoke(self, context, event):
...
 
+class BakeLightmaps(Operator):
+    bl_idname = "object.bake_lightmaps"
+    bl_label = "Bake Lightmaps"
+    bl_description = "Bake lightmaps of selected objects using the Cycles render engine and pack them into the .blend."
+    bl_options = {'REGISTER', 'UNDO'}
+
+    default_intensity: FloatProperty(name="Lightmaps Intensity",
+                                     default=3.14,
+                                     description="Multiplier for hubs on how to interpert the brightness of the image. Set this to 1.0 if you have set up the lightmaps manually and use a non-HDR format like png or jpg.")
```
vincentfretin, line 682: interpert -> interpret

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +670,174 @@ def invoke(self, context, event):
...
+                # The new layer is the last in the list, swap it for position 1
+                obj_uv_layers[1], obj_uv_layers[-1] = obj_uv_layers[-1], obj_uv_layers[1]
+
+            # The layer for the lightmap needs to be the active one before lightmap packing
+            obj_uv_layers.active = obj_uv_layers['UV1']
+
+        # run UV lightmap packing on all selected objects
+        bpy.ops.object.mode_set(mode='EDIT')
+        bpy.ops.mesh.select_all(action='SELECT')
+        # TODO: We need to warn the user at some place like the README that the uv_layer[1] gets completely overwritten if it is called 'UV1'
```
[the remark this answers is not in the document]
GottfriedHofmann: Added to README 👍.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,207 @@ def invoke(self, context, event):
...
+        print(material_object_associations.items())
+        # Set up the UV layer structure and auto-unwrap optimized for lightmaps
+        visited_objects = set()
+        for mat, obj_list in material_object_associations.items():
+            for ob in visited_objects:
+                if ob in obj_list:
+                    obj_list.remove(ob)
+            self.create_uv_layouts(context, obj_list)
+            for ob in obj_list:
+                visited_objects.add(ob)
```
vincentfretin, line 773: This visited_objects logic seems wrong. I spend 1h thinking about it, I hope my example will be clear.

We agree that an object can have several materials, each material with their own image texture, the lightmap uvmap is shared for those materials on an object.

Example:
obj1 with alu and glass materials
obj2 with alu
obj3 with glass
obj4 with rock

We handle alu material first, create uvmap for obj1 obj2 on same image texture for alu.
Then we handle glass material, we don't recreate the uvmap for obj1 because of your visited_objects, we create the uvmap for obj3. The faces for obj1 and obj3 may end up overlapping on the same image texture for glass.

What we need to do is create the uvmap for obj1 obj2 obj3 with the same smart uv project.
For example with alu, glass and rock materials, 
iterate over materials, iterating over alu objects, if an object has other materials, include all objects of those materials to do the uv smart project. Mark alu and glass as done. Continue iterating over materials, skip glass because it's done, then handle rock material.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,207 @@ def invoke(self, context, event):
...
+
+        # return to old settings
+        bake_settings = bake_settings_before
+        context.scene.cycles.samples = samples_tmp
+        context.scene.render.engine = render_engine_tmp
+
+        return {'FINISHED'}
+
+    def invoke(self, context, event):
+        # needed to get the dialoge with the intensity
```
vincentfretin, line 842: dialoge -> dialog

addons/io_hubs_addon/components/operators.py
GottfriedHofmann: You are totally right, it needs to work by material and not by object, should be adressed by https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/281/commits/a91a5b43f363b1558e9f551c34f62f05eb42f991

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,212 @@ def invoke(self, context, event):
...
+        # Set up the UV layer structure and auto-unwrap optimized for lightmaps
+        visited_materials = set()
+        for mat, obj_list in material_object_associations.items():
+            objs_to_uv_unwrap = []
+            if mat not in visited_materials:
+                # Several objects can share the same material so bundle them all
+                for ob in obj_list:
+                    for slot in ob.material_slots:
+                        if slot.material is not None:
+                            objs_to_uv_unwrap.extend(material_object_associations[slot.material])
```
vincentfretin, line 775: If I'm not mistaken, you may end up having an object several times in `objs_to_uv_unwrap`. You probably want to use a set and use `objs_to_uv_unwrap.update` here.

README.md
```diff
@@ -38,6 +38,10 @@ Note that for use in Hubs, you currently **MUST** use the second UV set, as Thre
 
 ![setting bake UV](https://user-images.githubusercontent.com/130735/83697782-b9e96b00-a5b4-11ea-986b-6690c69d8a3f.png)
 
+# Automatically baking Lightmaps
+
+To automatically create the node-setup needed to bake lightmaps and run baking on one step, select all objects you want to bake lightmaps for and got to `Object Properties > Hubs Lightmap Baker` and click on `Bake Lightmaps of selected objects`. **WARNING**: If a second UV layer is present on an object but it does not have a material with a `MOZ_lightmap` node, the UV layer will be overwritten!
```
Exairnous:
```suggestion
To automatically create the node-setup needed to bake lightmaps and run baking on one step, select all objects you want to bake lightmaps for and go to `Object Properties > Hubs Lightmap Baker` and click on `Bake Lightmaps of selected objects`. **WARNING**: If a second UV layer is present on an object but it does not have a material with a `MOZ_lightmap` node, the UV layer will be overwritten!
```

.vscode/settings.json
```diff
@@ -1,18 +1,19 @@
...
+  "python.analysis.extraPaths": [
+    "./fake_bpy_modules_3.3-20221006"
+  ],
+  "editor.defaultFormatter": "ms-python.autopep8",
+  "editor.formatOnSave": true,
+  "editor.formatOnSaveMode": "file",
+  "autopep8.args": [
+    "--exclude=models", "--max-line-length", "120", "--experimental"
+  ],
+  "cmake.configureOnOpen": false,
```
Exairnous: I don't think this change to the vscode settings is needed?

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,273 @@ def invoke(self, context, event):
...
+                    node.image.packed_files[0].filepath = new_filepath
+                    node.image.filepath_raw = new_filepath
+                    node.image.filepath = new_filepath
+
+                    # Remove file from temporary directory to de-clutter the system. Especially on windows the temporary directory is rarely purged.
+                    if os.path.exists(file_path):
+                        os.remove(file_path)
+
+                # return to old settings
+                bake_settings = self.bake_settings_before
```
Exairnous: This doesn't restore the old bake settings, they need to be handled one by one, the same as the render settings.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,273 @@ def invoke(self, context, event):
...
+                # filter mesh objects and others
+                mesh_objs, other_objs = [], []
+                for ob in selected_objects:
+                    (mesh_objs if ob.type == 'MESH' else other_objs).append(ob)
+                    # Remove all objects from selection so we can easily re-select subgroups later
+                    ob.select_set(False)
+
+                # The active object should be a MESH object
+                if context.active_object.type != 'MESH':
+                    context.view_layer.objects.active = mesh_objs[0]
```
Exairnous: Is this needed as well as the operator poll?

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,273 @@ def invoke(self, context, event):
...
+                    if len(obj.material_slots) >= 1:
+                        for slot in obj.material_slots:
+                            if slot.material is not None:
+                                mat = slot.material
+                                if mat not in material_object_associations:
+                                    material_object_associations[mat] = []
+                                material_object_associations[mat].append(obj)
+                    else:
+                        print("Object " + obj.name + " does not have material slots, removing from list of objects that will be unwrapped.")
+                        mesh_objs.remove(obj)
```
Exairnous: Is it a problem if the active object is removed?

Also, I think this should be shown to the user via the operator's report method in addition to printing to the console (which the user doesn't see).  This would also help the user understand why the bake failed if they try to bake a singular object that doesn't have materials as currently the only error message reported by Blender is that no valid objects are selected.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,273 @@ def invoke(self, context, event):
...
+    _timer = None
+    done = False
+    bake_started = False
+    cycles_samples_tmp = 1024
+    render_engine_tmp = 'CYCLES'
+    render_file_format_tmp = 'JPG'
+    bake_settings_before = False
+    lightmap_texture_nodes = []
+
+    image_type: EnumProperty(items=(('0', 'HDR', ''), ('1', 'JPEG', '')),
```
Exairnous: If you use the names as the identifiers, you can access them directly, e.g. `self.image_name` will return either `HDR` or `JPEG`.
```suggestion
    image_type: EnumProperty(items=(('HDR', 'HDR', ''), ('JPEG', 'JPEG', '')),
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,273 @@ def invoke(self, context, event):
...
+                            default=1024,
+                            description="The pixel resolution of the resulting lightmap.")
+    samples: IntProperty(name="Max Samples",
+                         default=1024,
+                         description="The number of samples to use for baking. Higher values reduce noise but take longer.")
+
+    def create_uv_layouts(self, context, mesh_objs):
+        # Sometimes the list can be empty, in that case do not unwrap but return directly
+        if len(mesh_objs) == 0:
+            return {'FINISHED'}
```
Exairnous: This won't exit the operator because it's the return value of a sub-call.  If you want to exit the operator, you'll need to have the `modal` method return `{'FINISHED'}` (although it would probably be better to have it return `{'CANCELLED'}` so that an undo step isn't created).  To do this, I would suggest returning a success/failure value from `create_uv_layouts` then checking that for failure and returning `{'CANCELLED'}` from the `modal` method if it's failed.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,273 @@ def invoke(self, context, event):
...
+            0.5, window=context.window)
+        context.window_manager.modal_handler_add(self)
+        return {"RUNNING_MODAL"}
+
+    def modal(self, context, event):
+        print("Running Modal")
+        if event.type == 'TIMER':
+            if not self.bake_started:
+                # Check selected objects
+                selected_objects = bpy.context.selected_objects
```
Exairnous: This isn't restored when the operator finishes.  It would be nice to leave the selection in the same state as when the operator is triggered (and the active object should be restored too).

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,273 @@ def invoke(self, context, event):
...
+                return {"FINISHED"}
+
+        return {'PASS_THROUGH'}
+
+    def invoke(self, context, event):
+        # needed to get the dialog with the intensity
+        return context.window_manager.invoke_props_dialog(self)
+
+    @classmethod
+    def poll(cls, context):
```
Exairnous, line 874: While not necessary, it might be nice to display in the tooltip the reason the operator is disabled with `cls.poll_message_set("Reason the poll failed")`.

Note: this isn't consistently done throughout the add-on, so definitely not a hard requirement.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,273 @@ def invoke(self, context, event):
...
+
+        lightmap_texture_node = mat_nodes.new(type="ShaderNodeTexImage")
+        lightmap_texture_node.location[0] -= 300
+
+        img = bpy.data.images.new('LightMap', self.resolution, self.resolution, alpha=False, float_buffer=True)
+        lightmap_texture_node.image = img
+        if bpy.app.version < (4, 0, 0):
+            lightmap_texture_node.image.colorspace_settings.name = "Linear"
+        else:
+            lightmap_texture_node.image.colorspace_settings.name = "Linear Rec.709"
```
Exairnous, line 909: I don't see any errors reported, but this also doesn't seem to fully work as the colorspace is displayed as sRGB when I go to check the image after baking for both HDR and JPEG.  If I watch closely it looks like it is initially set to Linear, but then it gets changed to sRGB when the bake finishes.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,273 @@ def invoke(self, context, event):
...
+                enum_items = self.properties.bl_rna.properties['image_type'].enum_items
+                self.image_type_name = enum_items.get(self.image_type).name
+                # Not sure whether this has any influence
+                bake_settings.image_settings.file_format = self.image_type_name
+                context.scene.render.image_settings.file_format = self.image_type_name
+
+                bpy.ops.object.bake('INVOKE_DEFAULT', type='DIFFUSE')
+                self.bake_started = True
+                return {"PASS_THROUGH"}
+            elif bpy.app.is_job_running('OBJECT_BAKE'):
```
Exairnous: I think this should be added to the poll as well so the user can't initiate another bake while one is already in progress.

Exairnous (changes requested):
Thank you for the changes!  Overall, this looks generally good, but there are some things I've left inline comments on.

.vscode/settings.json
GottfriedHofmann: No, not at all. It does not even change a thing.

addons/io_hubs_addon/components/operators.py
GottfriedHofmann: Thanks for catching this one, I have switched to the methods from reflection_probe.py, much cleaner now when restoring the settings.

addons/io_hubs_addon/components/operators.py
GottfriedHofmann: Well this should not be needed the operator poll should prevent this case.

addons/io_hubs_addon/components/operators.py
GottfriedHofmann: That is actually a really nice feature, thanks!

addons/io_hubs_addon/components/operators.py
GottfriedHofmann: Very good suggestion implemented in https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/281/commits/ece8ee2dab4aa6e9c5148d844fc01d974751fbab

addons/io_hubs_addon/components/operators.py
GottfriedHofmann: Thanks, addressed by https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/281/commits/773fb2c75cc97ce578ee4ef618a8ebed41cc200f

addons/io_hubs_addon/components/operators.py
GottfriedHofmann: In this case it is not a problem if the active object is removed (because it is only removed from selection, it stays active). https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/281/commits/773fb2c75cc97ce578ee4ef618a8ebed41cc200f adds an error report for the user.

addons/io_hubs_addon/components/operators.py
GottfriedHofmann: Moved the check up one level in https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/281/commits/773fb2c75cc97ce578ee4ef618a8ebed41cc200f

addons/io_hubs_addon/components/operators.py
GottfriedHofmann: Wow, there is actually a whole can of worms behind this one. The problem seems to have been introduced with the switch to non-blocking with a modal operator. Here are the issues I have encountered when investigating this:

- The packed images seem not to get saved, you have to prompt the save yourself. Blender will ask you about that when closing or loading another file but many users will probably just click it away, losing their bake.
- The images get saved to disk, but are completely black with all pixel values at 0.0. The image datablock loaded in Blender still has the color information, though.
- As mentioned the color space gets set to SRGB at the end.

Exairnous: I had some extra time at the dev meetup today, so I looked into this a little and I found that the image is trying to save before it's finished baking and that's why it's not getting the colorspace or being saved properly.  You'll need to add a handler for `bpy.app.handlers.object_bake_complete`, which will then trigger when the bake is finished so you can let the modal timer know when it's time to save the image (remember to remove the handler after it's been triggered).  Reflection probes do a similar thing with their baking, so you can use that as a reference.  Note: I'm not sure you need to reload the image after packing.

I need to focus mainly on the branding removal project/sprint right now (see the [Programming Team Roadmap](https://docs.google.com/document/d/1xRSHEgf4jE4SRsx2Z5BX_uXKbJ7KuCW3v0ntugA76io/edit?tab=t.0#heading=h.y5llc2g36m1n) and the [roadmaps policy](https://github.com/Hubs-Foundation/policies-procedures-guidelines-public/blob/main/roadmaps-policy.md) for more information on that), but if you can fix these last couple issues and get the baking PR ready, I'll review it as soon as the branding removal project/sprint is done.

If you have any questions about the handler stuff (or anything else), feel free to ask.

Exairnous: @GottfriedHofmann Although we're still in the sprint, it turns out we will be doing a larger release of the Blender add-on than expected, so if you can finish this before the release, I would love to include it.

GottfriedHofmann: @Exairnous  Thank you for the handler idea, it seems to work pretty well, check out my latest changes.

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
 
+class BakeLightmaps(Operator):
+    bl_idname = "object.bake_lightmaps"
+    bl_label = "Bake Lightmaps"
+    bl_description = "Bake lightmaps of selected objects using the Cycles render engine and pack them into the .blend."
+    bl_options = {'REGISTER', 'UNDO'}
+
+    # Global vars for modal operator
+    _timer = None
+    done = False
```
Exairnous, line 683:
```suggestion
    cancelled = False
    done = False
    errors = 0
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+                        if slot.material is not None:
+                            objs_to_uv_unwrap.update(material_object_associations[slot.material])
+                            visited_materials.add(slot.material)
+            print("Objects to UV unwrap: " + str(objs_to_uv_unwrap))
+        return objs_to_uv_unwrap
+
+    def execute(self, context):
+        # This function manages the handler that will run over and over.
+        # First reset state flags.
+        self.done = False
```
Exairnous, line 780:
```suggestion
        self.cancelled = False
        self.done = False
        self.errors = 0
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+                    if prop not in self.saved_props:
+                        self.saved_props[prop] = rgetattr(bpy.context, prop)
+                    rsetattr(bpy.context, prop, value)
+                # Start the bake process.
+                bpy.ops.object.bake('INVOKE_DEFAULT', type='DIFFUSE')
+                self.bake_started = True
+                # Remain in modal state while baking.
+                return {"PASS_THROUGH"}
+            # We only enter this block if the bake_complete_handler has flipped self.done to True.
+            elif self.done:
```
Exairnous:
```suggestion
            elif self.done or self.cancelled:
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+                # Pack all newly created or updated images
+                for node in self.lightmap_texture_nodes:
+                    file_path = bpy.path.abspath(f"{bpy.app.tempdir}/{node.image.name}.{self.image_type.lower()}")
+                    # node.image.save_render(file_path)
+                    node.image.filepath_raw = file_path
+                    node.image.file_format = self.image_type
+                    node.image.save()
```
Exairnous:
```suggestion
                    node.image.save_render(file_path)
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+                # context.selected_objects is not writable so we have to set the selection state on each individual object
+                for ob in self.selected_objects:
+                    ob.select_set(True)
+
+                # Clean up handlers and timer
+                if self.bake_complete_handler in bpy.app.handlers.object_bake_complete:
+                    bpy.app.handlers.object_bake_complete.remove(self.bake_complete_handler)
+                context.window_manager.event_timer_remove(self._timer)
+
+                self.report({'INFO'}, "Baking Hubs lightmaps completed successfully!")
```
Exairnous:
```suggestion
                if self.errors:
                    self.report({'ERROR'}, f"Baking Hubs lightmaps completed with {errors} errors!")
                else:
                    if self.done:
                        self.report({'INFO'}, "Baking Hubs lightmaps completed successfully!")
                    elif self.cancelled:
                        self.report({'INFO'}, "Baking Hubs lightmaps cancelled!")
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+
+    def get_lightmap_texture_nodes(self, material_object_associations):
+        # Check for the required nodes and set them up if not present
+        lightmap_texture_nodes = []
+        for mat in material_object_associations.keys():
+            mat_nodes = mat.node_tree.nodes
+            lightmap_nodes = [node for node in mat_nodes if node.bl_idname == 'moz_lightmap.node']
+            number_of_lightmap_nodes = len(lightmap_nodes)
+            if number_of_lightmap_nodes > 1:
+                print(str(number_of_lightmap_nodes) + " lightmap nodes in node tree of material " + mat.name + ". There should only be one!")
```
Exairnous:
```suggestion
                self.errors += 1
                self.report({'ERROR'}, f"{number_of_lightmap_nodes} lightmap nodes in node tree of material  {mat.name}. There should only be one!")
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+            elif len(lightmap_nodes) < 1:
+                lightmap_texture_nodes.append(self.setup_moz_lightmap_nodes(mat.node_tree))
+            else:
+                # TODO: Check wether all nodes are set up correctly. There is another pull request open for this. For now assume they are.
+                lightmap_nodes[0].intensity = self.default_intensity
+                # the image texture node needs to be the active one for baking, it is connected to the lightmap node so get it from there
+                lightmap_texture_node = lightmap_nodes[0].inputs[0].links[0].from_node
+                mat.node_tree.nodes.active = lightmap_texture_node
+                lightmap_texture_nodes.append(lightmap_texture_node)
```
Exairnous:
```suggestion
                # The image texture node needs to be the active one for baking and for Blender 5.0+ it must also be selected (see https://projects.blender.org/blender/blender/pulls/137389).  It is connected to the lightmap node so get it from there.
                try:
                    lightmap_texture_node = lightmap_nodes[0].inputs[0].links[0].from_node

                    if not hasattr(lightmap_texture_node, "image") or not lightmap_texture_node.image:
                        self.errors += 1
                        self.report({'ERROR'}, f"An invalid image texture node is connected to the lightmap node in {mat.name}")
                        continue

                    lightmap_texture_node.select = True
                    mat.node_tree.nodes.active = lightmap_texture_node
                    lightmap_texture_nodes.append(lightmap_texture_node)
                except IndexError:
                    self.errors += 1
                    self.report({'ERROR'}, f"No image texture node is connected to the lightmap node in {mat.name}")
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+    resolution: IntProperty(name="Lightmaps Resolution",
+                            default=1024,
+                            description="The pixel resolution of the resulting lightmap.")
+    samples: IntProperty(name="Max Samples",
+                         default=1024,
+                         description="The number of samples to use for baking. Higher values reduce noise but take longer.")
+    bake_margin: IntProperty(name="Bake Margin",
+                             default=4,
+                             description="How far the lightmaps extends the UV bounds. If you encounter black spots on the lightmap, try increasing this value.")
+
```
Exairnous, line 709:
```suggestion
    # Handler method to catch when bake is cancelled.
    def bake_cancel_handler(self, context, scene):
        self.cancelled = True

```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+            print("Objects to UV unwrap: " + str(objs_to_uv_unwrap))
+        return objs_to_uv_unwrap
+
+    def execute(self, context):
+        # This function manages the handler that will run over and over.
+        # First reset state flags.
+        self.done = False
+        self.bake_started = False
+        self.saved_props = {}
+        # Then attach the bake complete handler to Blender's app handlers.
```
Exairnous:
```suggestion
        # Then attach the bake cancelled/complete handlers to Blender's app handlers.
        if self.bake_cancel_handler not in bpy.app.handlers.object_bake_cancel:
            bpy.app.handlers.object_bake_cancel.append(self.bake_cancel_handler)
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+                        mesh_objs.remove(obj)
+
+                if len(mesh_objs) < 1:
+                    self.report(
+                        {'ERROR'}, f"No objects valid for baking in the current selection, aborting. \nObjects need materials assigned, otherwise they will not be baked.")
+                    # return to old selection state
+                    for ob in self.selected_objects:
+                        ob.select_set(True)
+
+                    # Clean up timer and bake complete handler
```
Exairnous:
```suggestion
                    # Clean up timer and bake cancelled/complete handlers
                    if self.bake_cancel_handler in bpy.app.handlers.object_bake_cancel:
                        bpy.app.handlers.object_bake_cancel.remove(self.bake_cancel_handler)
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+
+                # return to old settings
+                for prop in self.saved_props:
+                    rsetattr(bpy.context, prop, self.saved_props[prop])
+
+                # context.selected_objects is not writable so we have to set the selection state on each individual object
+                for ob in self.selected_objects:
+                    ob.select_set(True)
+
+                # Clean up handlers and timer
```
Exairnous, line 898:
```suggestion
                # Clean up handlers and timer
                if self.bake_cancel_handler in bpy.app.handlers.object_bake_cancel:
                    bpy.app.handlers.object_bake_cancel.remove(self.bake_cancel_handler)
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+            lightmap_texture_node.image.colorspace_settings.name = "Linear Rec.709"
+
+        UVmap_node = mat_nodes.new(type="ShaderNodeUVMap")
+        UVmap_node.uv_map = LIGHTMAP_LAYER_NAME
+        UVmap_node.location[0] -= 500
+
+        node_tree.links.new(UVmap_node.outputs['UV'], lightmap_texture_node.inputs['Vector'])
+        node_tree.links.new(lightmap_texture_node.outputs['Color'], lightmap_node.inputs['Lightmap'])
+
+        # the image texture node needs to be the active one for baking
```
Exairnous:
```suggestion
        # The image texture node needs to be the active one for baking and for Blender 5.0+ it must also be selected (see https://projects.blender.org/blender/blender/pulls/137389)
        lightmap_texture_node.select = True
```

addons/io_hubs_addon/components/operators.py
Exairnous: @GottfriedHofmann Thank you for the changes.  I've looked them over and I discovered that I forgot to mention a crucial part in solving the issue with the correct color space not getting saved/applied (I only remembered/figured it out when I noticed it wasn't working and compared your changes to the experiments I'd made).  I'm very sorry about this.

Anyway, the fix is simple; instead of setting the filepath and doing a regular save of the image, you need to do a `save_render` and pass the filepath (just like you had commented out above).  I also found that the bake getting cancelled needs to be handled as well, otherwise the modal operator will just continue indefinitely.

I've added the needed code as inline suggestions, along with a fix for Blender 5.0+ (it requires the node to also be selected, in addition to being active) and some additional error handling for if there is no image texture node connected or if the image texture node didn't have an image assigned to it and (what I think is) some improved error reporting.  The errors I handled were Python errors, so I figured it was better to handle them and prevent the chance that things could be left in an undefined state and cause problems for future bakes (and since I went that far, I figured it would be good to add the extra error messages too).

So, have a look over all of this, see what you think and apply what you like (or re-implement it another way) and I'll re-review once you've finished.

Exairnous (changes requested):
@GottfriedHofmann 
See my main reply in the inline comment thread on the color space not getting saved.

But since I need a main review comment to make GitHub happy, here is an overview of the inline suggestions I've left (fingers crossed that I've copied them all correctly and haven't missed anything):
* Handling bake cancels
* Gracefully handling additional errors in the node setup
* Improved error reporting
* Saving the image with the correct color space
* And supporting Blender 5.0+

addons/io_hubs_addon/components/operators.py
Exairnous: @GottfriedHofmann I may have spoken too soon about the color space being fixed.  I just tried saving it as a JPEG and it also saved in a linear color space (I'm pretty sure JPEG only supports sRGB).  This was with my fixes, so I'm not sure exactly what the issue is.  Could you look into this and double check, please?

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,307 @@ def invoke(self, context, event):
...
+        ]
+        if event.type == 'TIMER':
+            if not self.bake_started:
+                # Check selected objects and store them for later restoration
+                self.selected_objects = context.selected_objects
+
+                # filter mesh objects and others
+                mesh_objs, other_objs = [], []
+                for ob in self.selected_objects:
+                    (mesh_objs if ob.type == 'MESH' else other_objs).append(ob)
```
Exairnous: Oh, and is other_objs needed?  It looks to be essentially unused.

Note: at this point, I'm a little wary of making more changes, so if you'd prefer to leave this in (even if it doesn't do anything), that's fine with me.

Exairnous:
ROADMAP STATUS UPDATE: This PR has been moved to the short term section of the [Programming Team Roadmap](https://docs.google.com/document/d/1xRSHEgf4jE4SRsx2Z5BX_uXKbJ7KuCW3v0ntugA76io/edit?tab=t.0).

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,340 @@ def invoke(self, context, event):
...
+    errors = 0
+    bake_started = False
+    lightmap_texture_nodes = []
+    # properties to restore after operator is finished
+    selected_objects = []
+    saved_props = {}
+
+    image_type: EnumProperty(items=(('HDR', 'HDR', ''), ('JPEG', 'JPEG', '')),
+                             name="Image Type",
+                             description="The image type used to store the lighting information. HDR supports full 32 bit lighting information but can lead to huge files. JPG only stores limited lighting information but results in much smaller files.")
```
Exairnous, line 693: Add in note that this only affects new images that are created.  Also changes JPG -> JPEG to match the options offered.
```suggestion
                             description="The image type used to store the lighting information (only applies to new image nodes added in this bake). HDR supports full 32 bit lighting information but can lead to huge files. JPEG only stores limited lighting information but results in much smaller files")
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,340 @@ def invoke(self, context, event):
...
+    default_intensity: FloatProperty(name="Lightmaps Intensity",
+                                     default=3.14,
+                                     description="Multiplier for hubs on how to interpret the brightness of the image. Set this to 1.0 if you use JPG or another non-HDR format.")
+    resolution: IntProperty(name="Lightmaps Resolution",
+                            default=1024,
+                            description="The pixel resolution of the resulting lightmap.")
```
Exairnous, line 699: Add in note that this only affects new images that are created.
```suggestion
                            description="The pixel resolution of the resulting lightmap (only applies to new image nodes added in this bake).")
```

README.md
```diff
@@ -38,6 +38,10 @@ Note that for use in Hubs, you currently **MUST** use the second UV set, as Thre
 
 ![setting bake UV](https://user-images.githubusercontent.com/130735/83697782-b9e96b00-a5b4-11ea-986b-6690c69d8a3f.png)
 
+# Automatically baking Lightmaps
+
+To automatically create the node-setup needed to bake lightmaps and run baking on one step, select all objects you want to bake lightmaps for and go to `Object Properties > Hubs Lightmap Baker` and click on `Bake Lightmaps of selected objects`. **WARNING**: If a second UV layer is present on an object but it does not have a material with a `MOZ_lightmap` node, the UV layer will be overwritten!
```
Exairnous, line 43: I believe the second UV layer is always overwritten, though if it doesn't have the expected name, it's moved to/preserved as a new layer, so what about giving this warning?
```suggestion
To automatically create the node-setup needed to bake lightmaps and run baking on one step, select all objects you want to bake lightmaps for and go to `Object Properties > Hubs Lightmap Baker` and click on `Bake Lightmaps of selected objects`. **WARNING**: If a second UV layer is present on an object it will be overwritten (or moved to a new layer if its name isn't LightmapUV)!
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,340 @@ def invoke(self, context, event):
...
+
+                # Clean up handlers and timer
+                if self.bake_cancel_handler in bpy.app.handlers.object_bake_cancel:
+                    bpy.app.handlers.object_bake_cancel.remove(self.bake_cancel_handler)
+                if self.bake_complete_handler in bpy.app.handlers.object_bake_complete:
+                    bpy.app.handlers.object_bake_complete.remove(self.bake_complete_handler)
+                context.window_manager.event_timer_remove(self._timer)
+
+                if self.errors:
+                    self.report({'ERROR'}, f"Baking Hubs lightmaps completed with {errors} errors!")
```
Exairnous, line 906: I mistyped the variable name in my initial suggestion.  Sorry!  It should have been `self.errors`.

I also think the user should be told to check the Info editor as they may not know where to look for more information on the errors, so I've included that in this suggestion too.
```suggestion
                    self.report({'ERROR'}, f"Baking Hubs lightmaps completed with {self.errors} errors!  Check the Info editor for more information")
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,340 @@ def invoke(self, context, event):
...
+                                mat = slot.material
+                                if mat not in material_object_associations:
+                                    material_object_associations[mat] = []
+                                material_object_associations[mat].append(obj)
+                    else:
+                        self.report(
+                            {'INFO'}, f"Object {obj.name} does not have material slots, removing it from list of objects that will be unwrapped and baked.")
+                        mesh_objs.remove(obj)
+
+                if len(mesh_objs) < 1:
```
Exairnous, line 838: Objects can have material slots, but no materials assigned to them, so checking `material_object_associations` should give a more accurate result of whether there are valid objects for baking.

```suggestion
                if len(material_object_associations) < 1:
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,340 @@ def invoke(self, context, event):
...
+                for (prop, value) in overrides:
+                    if prop not in self.saved_props:
+                        self.saved_props[prop] = rgetattr(bpy.context, prop)
+                    rsetattr(bpy.context, prop, value)
+                # Start the bake process.
+                bpy.ops.object.bake('INVOKE_DEFAULT', type='DIFFUSE')
+                self.bake_started = True
+                # Remain in modal state while baking.
+                return {"PASS_THROUGH"}
+            # We only enter this block if the bake_complete_handler has flipped self.done to True.
```
Exairnous, line 871: Update comment to include mention of cancelled bakes.
```suggestion
            # We only enter this block if the bake_complete_handler/bake_cancel_handler has flipped self.done/self.cancelled to True.
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,340 @@ def invoke(self, context, event):
...
+            elif self.done or self.cancelled:
+                # Pack all newly created or updated images
+                for node in self.lightmap_texture_nodes:
+                    file_path = bpy.path.abspath(f"{bpy.app.tempdir}/{node.image.name}.{self.image_type.lower()}")
+                    node.image.save_render(file_path)
+                    node.image.pack()
+                    node.image.reload()
```
Exairnous, line 878: During my testing, I found that JPEG files weren't packing/unpacking correctly (the contents of the file were reported as EXR, not JPEG).  I guess the generated version of the image (with all the raw render data) was getting packed instead of the saved out version (that had been converted to JPEG).  This change appears to fix it (before the change, unpacking the JPEG of a bake of the default scene/cube resulted in a 12MiB file and after the change the file is only 27.5KiB).

Curiously, I think HDR files were getting packed/unpacked fine without this change, but the change doesn't break anything for them.

```suggestion
                    node.image.source = 'FILE'
                    node.image.filepath = file_path
                    node.image.filepath_raw = file_path
                    node.image.reload()
                    node.image.pack()
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,340 @@ def invoke(self, context, event):
...
+                if self.errors:
+                    self.report({'ERROR'}, f"Baking Hubs lightmaps completed with {errors} errors!")
+                else:
+                    if self.done:
+                        self.report({'INFO'}, "Baking Hubs lightmaps completed successfully!")
+                    elif self.cancelled:
+                        self.report({'INFO'}, "Baking Hubs lightmaps cancelled!")
+                return {"FINISHED"}
+
+        # If self.bake_started is True but self.done is false, just wait.
```
Exairnous, line 914: Update comment to include mention of cancelled bakes.
```suggestion
        # If self.bake_started is True but self.done and self.cancelled is false, just wait.
```

addons/io_hubs_addon/components/operators.py
```diff
@@ -670,6 +671,340 @@ def invoke(self, context, event):
...
+    # properties to restore after operator is finished
+    selected_objects = []
+    saved_props = {}
+
+    image_type: EnumProperty(items=(('HDR', 'HDR', ''), ('JPEG', 'JPEG', '')),
+                             name="Image Type",
+                             description="The image type used to store the lighting information. HDR supports full 32 bit lighting information but can lead to huge files. JPG only stores limited lighting information but results in much smaller files.")
+    default_intensity: FloatProperty(name="Lightmaps Intensity",
+                                     default=3.14,
+                                     description="Multiplier for hubs on how to interpret the brightness of the image. Set this to 1.0 if you use JPG or another non-HDR format.")
```
Exairnous, line 696: Changes JPG -> JPEG to match the options offered.
```suggestion
                                     description="Multiplier for hubs on how to interpret the brightness of the image. Set this to 1.0 if you use JPEG or another non-HDR format.")
```

Exairnous (changes requested):
@GottfriedHofmann 
I found a few more things.  Hopefully this is the last of it. 🤞 
Aside from my inline comments/suggestions, I believe it needs a run of format.py

Summary of inline comments/suggestions
* Tooltip updates.
* Readme warning update.
* Error message update and fix of typo that caused a Python error.
* Comment updates.
* More fixes for saving JPEGs.

Exairnous:
I'm not mentioning this as a review comment because this shouldn't be addressed in this PR, but I wanted to note that I noticed that the setting restoration code doesn't handle all situations, e.g. it'll fail if the user has it set to render out movie files/FFMPEG, as well as with some options related to image types (though this is potentially dependent on Blender version).  The same limitation is present for baking reflection probes, so we'll want a unified fix for this that both the lightmap baking and reflection probe baking can use, but this should definitely be its own PR.

If you're interested, you're welcome to file a bug report on this and/or work on a fix; however, if you've had enough of baking related stuff, I totally understand (and as I wouldn't regard this as a critical issue, it would need to wait until the sprint is finished to move forward, unless maybe if it was a super simple fix, but I don't think that's likely).

I did find this BlenderArtists post that might serve as a useful reference for a fix, so I'm including the link for posterity.
https://blenderartists.org/t/save-render-result-with-custom-file-format-and-settings/1194782/6

PR #282, opened 2024-05-02
Scene grabbables support

keianhzo:
This PR adds support for a grabbable component that will make that object grabbable in Hubs (https://github.com/mozilla/hubs/pull/6487). The object can be made capturable by a media frame if the media frame collision bitmask is enabled.

Ammo shape has also been deprecated (hidden) as part of this as it can be replaced by the physics shape component.

See https://github.com/MozillaReality/blender-gltf-behavior-graph/pull/16


keianhzo:
@rawnsley In case you want to test/review

addons/io_hubs_addon/components/definitions/physics_shape.py
```diff
@@ -0,0 +1,118 @@
...
+        props = super().gather(export_settings, object)
+        props['offset'] = {
+            'x': self.offset[0],
+            'y': self.offset[2] if export_settings['gltf_yup'] else self.offset[1],
+            'z': self.offset[1] if export_settings['gltf_yup'] else self.offset[2],
+        }
+        props['halfExtents'] = {
+            'x': self.halfExtents[0],
+            'y': self.halfExtents[2] if export_settings['gltf_yup'] else self.halfExtents[1],
+            'z': self.halfExtents[1] if export_settings['gltf_yup'] else self.halfExtents[2],
```
Exairnous, line 101: Nit: I believe self.offset and self.halfExtents XYZ values could be specified by .x .y .z instead of by index, which would make things a little clearer.

addons/io_hubs_addon/components/definitions/rigid_body.py
```diff
@@ -0,0 +1,168 @@
...
+        props['angularFactor'] = [
+            self.angularFactor[0],
+            self.angularFactor[2] if export_settings['gltf_yup'] else self.angularFactor[1],
+            self.angularFactor[1] if export_settings['gltf_yup'] else self.angularFactor[2],
+        ]
+        props['gravity'] = [
+            self.gravity[0],
+            self.gravity[2] if export_settings['gltf_yup'] else self.gravity[1],
+            self.gravity[1] if export_settings['gltf_yup'] else self.gravity[2],
+        ]
```
Exairnous, line 121: Nit: I believe self.angularFactor and self.gravity XYZ values could be specified by .x .y .z instead of by index, which would make things a little clearer.

addons/io_hubs_addon/components/definitions/rigid_body.py
```diff
@@ -0,0 +1,168 @@
...
+        if instance_version <= (1, 0, 0):
+            migration_occurred = True
+
+            angularFactor = self.angularFactor.copy()
+            angularFactor = Vector((angularFactor.x, angularFactor.z, angularFactor.y))
+            self.angularFactor = angularFactor
+
+            # If the gravity property has not changed, it will take the default value so we don't swizzle
+            # in that case. Is there any way of checking if the property has been modified?
+            if self.gravity != Vector((0.0, 0.0, -9.8)):
```
Exairnous, line 163: To check if the property has been modified you should be able to do this:
`if self.get('gravity') is not None:`

addons/io_hubs_addon/debugger.py
```diff
@@ -976,7 +976,7 @@ class HubsSceneDebuggerRoomExportPrefs(bpy.types.PropertyGroup):
     avatar_to_viewport: bpy.props.BoolProperty(
         name='Spawn using viewport transform',
         description='Spawn the avatar in the current viewport camera position/rotation',
-        default=True, options=set())
+        default=False, options=set())
```
Exairnous, line 979: Why is the 'Spawn the avatar in the current viewport camera position/rotation' property being changed to default to off?

Exairnous (changes requested):
Sorry it took so long to get to this.  I've left some inline comments, but I also noticed that:
* Grabbable networked-object-properties components aren't being migrated to networked-transform components.
* The rigid-body component doesn't import properly.
* And ammo shapes can't be removed.

addons/io_hubs_addon/components/definitions/rigid_body.py
keianhzo: Oh cool, I didn't know that. Thanks.

addons/io_hubs_addon/debugger.py
keianhzo: I found it annoying that it was enabled by default as most of the time I don't want that so I just disabled it as I was working on this. I can move it to another PR.

PR #284, opened 2024-05-08
Isolate dependencies from other add-ons

keianhzo:
This PR isolates selenium from the Blender python packages to avoid conflicts. We still do it using pip but using a specific target directory and now the modules are loaded in isolation using importlib.

### DIFF
--- a/addons/io_hubs_addon/preferences.py
+++ b/addons/io_hubs_addon/preferences.py
@@ -51,7 +51,7 @@ class InstallDepsOperator(bpy.types.Operator):
             dep = f'{self.dep_config.name}=={self.dep_config.version}'
 
         result = subprocess.run(
-            [sys.executable, '-m', 'pip', 'install', dep,
+            [sys.executable, '-m', 'pip', 'install', '--upgrade', dep,
              '-t', get_or_create_deps_path(self.dep_config.name)],
             capture_output=True, text=True, input="y")
         failed = False
--- a/addons/io_hubs_addon/preferences.py
+++ b/addons/io_hubs_addon/preferences.py
@@ -1,6 +1,6 @@
 import bpy
 from bpy.types import AddonPreferences, Context
-from bpy.props import IntProperty, StringProperty, EnumProperty, BoolProperty, CollectionProperty, PointerProperty
+from bpy.props import IntProperty, StringProperty, EnumProperty, BoolProperty, PointerProperty
 from .utils import get_addon_package, is_module_available, get_browser_profile_directory
 import platform
 from os.path import join, dirname, realpath
@@ -72,11 +72,9 @@ class InstallDepsOperator(bpy.types.Operator):
 class UninstallDepsOperator(bpy.types.Operator):
     bl_idname = "pref.hubs_prefs_uninstall_dep"
     bl_label = "Uninstall a python dependency through pip"
-    bl_property = "dep_names"
     bl_options = {'REGISTER', 'UNDO'}
 
     dep_config: PointerProperty(type=DepsProperty)
-    force: BoolProperty(default=False)
 
     def execute(self, context):
         from .utils import get_or_create_deps_path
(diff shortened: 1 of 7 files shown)


addons/io_hubs_addon/preferences.py
```diff
@@ -88,52 +75,14 @@ class UninstallDepsOperator(bpy.types.Operator):
     bl_property = "dep_names"
     bl_options = {'REGISTER', 'UNDO'}
 
-    dep_names: CollectionProperty(type=DepsProperty)
+    dep_config: PointerProperty(type=DepsProperty)
     force: BoolProperty(default=False)
```
Exairnous, line 79: I think the force property can be removed now?

addons/io_hubs_addon/preferences.py
```diff
@@ -88,52 +75,14 @@ class UninstallDepsOperator(bpy.types.Operator):
     bl_property = "dep_names"
```
Exairnous, line 75: I'm not actually sure whether this bl_property specification does anything here, but it shows an error in a terminal with the changes.

Exairnous (commented):
This looks like a nice improvement.  I have a few questions though:
* Why have the install/uninstall deps operators been changed to only work with a single module?
* Why did you create a dependencies folder with an `__init__.py` instead of just creating dependencies.py?
* Why has the check to ensure pip is present been removed from the install deps operator?

Exairnous:
Oh, also, I think this will leave the old selenium module installed.  It would be nice to remove it, if that's not too hard.

addons/io_hubs_addon/preferences.py
keianhzo: I don't remember what is that doing there but I tested and removing it doesn't seem to affect so I've removed it.

keianhzo:
> Oh, also, I think this will leave the old selenium module installed. It would be nice to remove it, if that's not too hard.

Yeah that's a problem with the way we were doing things before we can't really remove it as we don't know if any other add-on might have installed or is using it.

Exairnous:
Thank you for the updates.

> Yeah that's a problem with the way we were doing things before we can't really remove it as we don't know if any other add-on might have installed or is using it.

Good point.  That makes sense.

Any update on my questions from above?
> Why have the install/uninstall deps operators been changed to only work with a single module?
> Why did you create a dependencies folder with an __init__.py instead of just creating dependencies.py?
> Why has the check to ensure pip is present been removed from the install deps operator?

keianhzo:
> Why have the install/uninstall deps operators been changed to only work with a single module?
We were only using these operators to install one dependency so I simplified the code to just get one dep.

> Why did you create a dependencies folder with an init.py instead of just creating dependencies.py?
No strong reason, I guess it makes it simple to do `from .dependencies import get_XXX`

>  Why has the check to ensure pip is present been removed from the install deps operator?
I think we can safely assume that pip is always present in Blender installs

Exairnous (changes requested):
> We were only using these operators to install one dependency so I simplified the code to just get one dep.

I figured it might be nice to have one button to install all dependencies and that supporting multiple would future proof that, but one button per dependency works too and is simpler.  So, I guess I don't have a strong opinion either way.

> No strong reason, I guess it makes it simple to do from .dependencies import get_XXX

I believe you can do that with a regular .py file, but this works fine too; just seems a little weird to me, that's all.

> I think we can safely assume that pip is always present in Blender installs

I just tried with Blender-3.3.2 on my system and got this error:
```
01   Dependencies install has failed
02   /home/<user>/Workshop/CustomApps/blender-3.3.2/3.3/python/bin/python3.10: No module named pip
```
So I think ensure pip needs to remain.  Some people recommend upgrading pip as well, but I'm not entirely sure if we need to: https://blender.stackexchange.com/questions/56011/how-to-install-pip-for-blenders-bundled-python

addons/io_hubs_addon/preferences.py
```diff
@@ -39,37 +39,24 @@ class InstallDepsOperator(bpy.types.Operator):
...
```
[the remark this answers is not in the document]
keianhzo: Updated

keianhzo:
I've addressed the feedback and added back pip and added pip upgrade.

Exairnous (approved):
LGTM.

PR #286, opened 2024-05-10
Prevent data corruption during a migration if components are removed.

Exairnous:
Depending on the order of the components, data corruption can occur if components are removed in the middle of a migration, e.g. like what is done in the grababble component's new migration.  This patch skips over components that have been removed and so doesn't try to migrate them, which should prevent any chance of data corruption.


keianhzo (approved):
LGTM

PR #287, opened 2024-05-10
Improve debugger texts

keianhzo:
Improve the Secne debugger buttons texts and descriptions for clarity.

### DIFF
--- a/addons/io_hubs_addon/debugger.py
+++ b/addons/io_hubs_addon/debugger.py
@@ -614,8 +614,8 @@ class HUBS_UL_ToolsSceneDebuggerRooms(bpy.types.UIList):
 
 class HubsPublishSceneOperator(bpy.types.Operator):
     bl_idname = "hubs_scene.publish_scene"
-    bl_label = "Scene Manager"
-    bl_description = "Publish scene"
+    bl_label = "Publish"
+    bl_description = "Publish current Blender scene"
     bl_options = {'REGISTER', 'UNDO'}
 
     @classmethod
@@ -684,7 +684,7 @@ class HubsPublishSceneOperator(bpy.types.Operator):
 class HubsUpdateSceneOperator(bpy.types.Operator):
     bl_idname = "hubs_scene.update_scene"
     bl_label = "Update"
-    bl_description = "Update selected scene"
+    bl_description = "Updates the selected scene with the Blender scene"
     bl_options = {'REGISTER', 'UNDO'}
 
     @classmethod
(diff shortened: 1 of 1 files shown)


addons/io_hubs_addon/debugger.py
```diff
@@ -53,7 +53,7 @@ def is_room_set(context):
 class HubsUpdateRoomOperator(bpy.types.Operator):
     bl_idname = "hubs_scene.update_room"
     bl_label = "View Scene"
-    bl_description = "Update room"
+    bl_description = "Updates the currently opened room scene with the Blender scene"
```
Exairnous, line 56: This should use the description classmethod so that the text can be changed depending on whether the scene is being updated or spawned as an object.

Something like this:
```
@classmethod
def description(cls, context, properties):
    is_scene_update = context.scene.hubs_scene_debugger_room_create_prefs.debugLocalScene
    if hubs_session.is_alive():
        room_params = hubs_session.room_params
        is_scene_update = "debugLocalScene" in room_params

    if is_scene_update:
        return "Updates the currently opened room scene with the Blender scene"
    else:
        return "Spawns the Blender scene in the currently opened room as an object"
```
https://docs.blender.org/api/current/bpy.types.Operator.html#bpy.types.Operator.description

addons/io_hubs_addon/debugger.py
```diff
@@ -122,8 +122,8 @@ def execute(self, context):
 
 class HubsCreateRoomOperator(bpy.types.Operator):
     bl_idname = "hubs_scene.create_room"
-    bl_label = "Create Room"
-    bl_description = "Create room"
+    bl_label = "Create a new room"
+    bl_description = "Creates a new room in the selected instance"
```
Exairnous, line 126: Maybe some additional info, like this: 
```suggestion
    bl_description = "Creates a new room in the selected instance and opens it in the browser selected in the add-on preferences.  The specified room flags will be applied"
```

addons/io_hubs_addon/debugger.py
```diff
@@ -153,8 +153,8 @@ def execute(self, context):
 
 class HubsOpenRoomOperator(bpy.types.Operator):
     bl_idname = "hubs_scene.open_room"
-    bl_label = "Open Room"
-    bl_description = "Open room"
+    bl_label = "Open selected room"
+    bl_description = "Opens the selected room"
```
Exairnous, line 157: Maybe some additional info, like this:
```suggestion
    bl_description = "Opens the selected room in the browser selected in the add-on preferences.  The specified room flags will be applied"
```

Exairnous (commented):
Looks pretty good.  I left a couple potential improvements in in-line comments.

The buttons in the publish section could also probably benefit from a similar treatment.

addons/io_hubs_addon/debugger.py
keianhzo: Oh cool, I dind't know this

addons/io_hubs_addon/debugger.py
keianhzo: Updated

addons/io_hubs_addon/debugger.py
keianhzo: Updated

keianhzo:
I addressed the feedback and I improved texts for the publish buttons too.

Exairnous:
Looks good.  Thanks!

PR #290, opened 2024-05-23
Support for user components directories

keianhzo:
This PR adds support for custom user components directories.

The user can add multiple directories and the components found in those directories will be loaded at add-on load time. 

We don't do hot reloading as it can get quite complex quickly and Blender already requires reload for it's own scripts paths AFAIK.

Required by https://github.com/MozillaReality/hubs-postprocessing-addon
Closes https://github.com/MozillaReality/hubs-blender-exporter/issues/288

### DIFF
--- a/addons/io_hubs_addon/components/components_registry.py
+++ b/addons/io_hubs_addon/components/components_registry.py
@@ -34,6 +34,35 @@ def get_components_in_dir(dir):
     return sorted(components)
 
 
+def get_user_component_names():
+    component_names = []
+    from ..preferences import get_addon_pref
+    addon_prefs = get_addon_pref(bpy.context)
+    for _, entry in enumerate(addon_prefs.user_components_paths):
+        if entry.path:
+            component_names = get_components_in_dir(entry.path)
+    return component_names
+
+
+def get_user_component_definitions():
+    modules = []
+    component_names = get_user_component_names()
+    for name in component_names:
+        from ..preferences import get_addon_pref
+        addon_prefs = get_addon_pref(bpy.context)
+        for _, entry in enumerate(addon_prefs.user_components_paths):
+            file_path = os.path.join(entry.path, name + ".py")
+            try:
+                spec = importlib.util.spec_from_file_location(name, file_path)
+                mod = importlib.util.module_from_spec(spec)
+                spec.loader.exec_module(mod)
+                modules.append(mod)
+
+            except Exception as e:
+                print(f'Failed import of component {name}', e)
+    return modules
+
+
 def get_component_definitions():
     components_dir = join(dirname(realpath(__file__)), "definitions")
     component_module_names = get_components_in_dir(components_dir)
@@ -113,6 +142,28 @@ def unregister_component(component_class):
         print(f"Component unregistered: {component_class.get_name()}")
 
 
+def load_user_components():
+    global __components_registry
+    for module in get_user_component_definitions():
+        for _, member in inspect.getmembers(module):
+            if inspect.isclass(member) and issubclass(member, HubsComponent) and module.__name__ == member.__module__:
+                if hasattr(module, 'register_module'):
+                    module.register_module()
+                register_component(member)
+                __components_registry[member.get_name()] = member
+
+
+def unload_user_components():
+    global __components_registry
+    for _, component_class in __components_registry.items():
+        for module_name in get_user_component_names():
+            if module_name == component_class.get_name():
+                unregister_component(component_class)
+    for module in get_user_component_definitions():
+        if hasattr(module, 'unregister_module'):
+            module.unregister_module()
+
+
 def load_components_registry():
     """Recurse in the components directory to build the components registry"""
     global __components_registry
@@ -124,6 +175,8 @@ def load_components_registry():
                     module.register_module()
                 register_component(member)
                 __components_registry[member.get_name()] = member
+
+    load_user_components()
 
 
 def unload_components_registry():
--- a/addons/io_hubs_addon/components/components_registry.py
+++ b/addons/io_hubs_addon/components/components_registry.py
@@ -38,28 +38,36 @@ def get_user_component_names():
     component_names = []
     from ..preferences import get_addon_pref
     addon_prefs = get_addon_pref(bpy.context)
-    for _, entry in enumerate(addon_prefs.user_components_paths):
+    for entry in addon_prefs.user_components_paths:
         if entry.path and os.path.isdir(entry.path):
-            component_names = get_components_in_dir(entry.path)
+            component_names.append(get_components_in_dir(entry.path))
     return component_names
+
+
+def get_user_component_paths():
+    component_paths = []
+    from ..preferences import get_addon_pref
+    addon_prefs = get_addon_pref(bpy.context)
+    for entry in addon_prefs.user_components_paths:
+        if entry.path and os.path.isdir(entry.path):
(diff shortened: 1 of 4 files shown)


addons/io_hubs_addon/components/components_registry.py
```diff
@@ -34,6 +34,35 @@ def get_components_in_dir(dir):
     return sorted(components)
 
 
+def get_user_component_names():
+    component_names = []
+    from ..preferences import get_addon_pref
+    addon_prefs = get_addon_pref(bpy.context)
+    for _, entry in enumerate(addon_prefs.user_components_paths):
+        if entry.path:
```
Exairnous, line 42: I think a check should be added here to make sure the path exists, otherwise the whole add-on can be broken.
```suggestion
        if entry.path and os.path.isdir(entry.path):
```

addons/io_hubs_addon/components/components_registry.py
```diff
@@ -113,6 +142,28 @@ def unregister_component(component_class):
...
 
+def load_user_components():
+    global __components_registry
+    for module in get_user_component_definitions():
+        for _, member in inspect.getmembers(module):
+            if inspect.isclass(member) and issubclass(member, HubsComponent) and module.__name__ == member.__module__:
+                if hasattr(module, 'register_module'):
+                    module.register_module()
+                register_component(member)
+                __components_registry[member.get_name()] = member
```
Exairnous, line 153: Currently, if there are any bugs in one of the user definitions it will break the whole add-on, so I think this section should be wrapped in a try/except block.  Something like this:
```suggestion
                try:
                    if hasattr(module, 'register_module'):
                        module.register_module()
                    register_component(member)
                    __components_registry[member.get_name()] = member
                except Exception:
                    import traceback
                    traceback.print_exc()
```

addons/io_hubs_addon/components/types.py
```diff
@@ -23,6 +23,7 @@ class Category(Enum):
     MISC = 'Misc'
     LIGHTS = 'Lights'
     MEDIA = 'Media'
+    USER = 'User'
```
Exairnous, line 26: The new `USER` type doesn't seem to be used anywhere currently.  Is it just another option, or was it supposed to be automatically assigned to user definitions somehow?

addons/io_hubs_addon/preferences.py
```diff
@@ -165,6 +165,101 @@ def execute(self, context):
         return {'FINISHED'}
 
 
+def reload_user_components():
+    from .components.components_registry import unload_user_components, load_user_components
+    unload_user_components()
+    load_user_components()
```
Exairnous, line 171: This reload_user_components function doesn't seem to be used anywhere.  Is there supposed to be a refresh button in the preferences that calls this?  (Currently there doesn't seem to be any way to pick up the definitions from a new folder without saving the preferences and restarting Blender.)

addons/io_hubs_addon/preferences.py
```diff
@@ -165,6 +165,101 @@ def execute(self, context):
...
+
+
+class HubsUserComponentsPath(bpy.types.PropertyGroup):
+    name: StringProperty(
+        name='User components path entry name',
+        description='The user components path entry name.',
```
Exairnous, line 177: I think this description could be improved.  How about something like:
```suggestion
        description='An optional, user defined label to allow quick discernment between different user component definition directories',
```

addons/io_hubs_addon/preferences.py
```diff
@@ -165,6 +165,101 @@ def execute(self, context):
...
+    )
+    path: StringProperty(
+        name='User components path path',
+        description='The user components path path. You can copy external components here.',
```
Exairnous, line 181: I think this description could be improved.  How about something like:
```suggestion
        description='The path to a user defined component definitions directory.  You can copy external components here and they will be loaded automatically',
```

addons/io_hubs_addon/preferences.py
```diff
@@ -165,6 +165,101 @@ def execute(self, context):
...
+    )
+
+
+class HubsUserComponentsPathAdd(bpy.types.Operator):
+    bl_idname = "hubs_preferences.add_user_components_path"
+    bl_label = "Add user components path"
+    bl_description = "Adds a new component path entry"
+    bl_options = {'REGISTER', 'UNDO'}
+
+    path: bpy.props.StringProperty(name="User Components Path", default="")
```
Exairnous, line 192: It looks like this `path` property is never passed anything and can be removed.

addons/io_hubs_addon/components/components_registry.py
```diff
@@ -125,6 +176,12 @@ def load_components_registry():
                 register_component(member)
                 __components_registry[member.get_name()] = member
 
+    # Preferences are not accessible until the add-on is enabled so we need to check that before loading the user
+    # components as they depend on the preferences.
+    from ..utils import is_addon_enabled
+    if is_addon_enabled():
+        load_user_components()
```
Exairnous, line 183: In which cases is the add-on not enabled enough when trying to load user components?  Is this related to running tests?
Also, I think this only gets called once, so if this check fails then I think the user components would never get loaded?

Exairnous (changes requested):
Thanks for this!
There were a few issues that I found that I think should be addressed, but I think they should be easy to solve.  I've left inline comments for them.

I also noticed in a couple places you're using `enumerate` but discarding the index.  Is there a reason why you aren't just looping over stuff regularly in those places?

addons/io_hubs_addon/components/components_registry.py
keianhzo: Added

addons/io_hubs_addon/components/components_registry.py
keianhzo: Added

addons/io_hubs_addon/components/components_registry.py
keianhzo: Yes, when running tests the add-on is added but not enabled so the access to preferences fails so that's the case this is trying to fix.

When the add-on is disabled preferences ar lost anyway so if the add-on is not enabled you will need to add the user paths and restart anyways.

addons/io_hubs_addon/components/types.py
keianhzo: It's only used in external components like here: https://github.com/Hubs-Foundation/hubs-postprocessing-addon/blob/main/blender/post_processing_effects.py

addons/io_hubs_addon/preferences.py
keianhzo: Added

addons/io_hubs_addon/preferences.py
keianhzo: Updated

addons/io_hubs_addon/preferences.py
keianhzo: Ops, yep, leftover. Remving...

Exairnous:
Oh, I just realized, when add/removing directories, the preferences should be marked dirty:
`context.preferences.is_dirty = True`

Exairnous (approved):
The dirty state of the preferences isn't updating when the fields are modified, which is weird.  I suppose you could add a lambda to the update argument for the string properties and manually mark them dirty, but not sure why it's not doing it automatically.  Anyway, I'll leave it up to you on whether you want to support this right now or not.

LGTM.  Thank you for the changes.

PR #295, opened 2024-06-05
Append version to zip output

keianhzo:
Append a version number to the end of the output zip artifact to make it easier to identify it.


Exairnous (approved):
LGTM.

PR #300, opened 2024-07-23
Fix type mismatches when assigning properties on import.

Exairnous:
Round numbers when assigning to integer properties on import to avoid errors relating to floating point inaccuracies, i.e. the property value should be an int but it's actually a float.


Exairnous:
Closing in favor of #302

Issue #308, opened 2024-07-30
Install Selenium in first click fails


hrithikwins:
**Description**
The Selenium module installation fails on first attempt 

**To Reproduce**
1. open blender
2. remove selenium completely
3. remove the blender addon
4. restart blender
5. install the addon
6. it'll show a 'Install selenium' dependency button
7. after clicking on it, it fails with an error for pip
8. restart blender
9. install the dependency again, it works

**Expected behavior**
Should have installed selenium dependency in first click

**Screenshots**
<img width="536" alt="image" src="https://github.com/user-attachments/assets/f1640f19-d428-44bc-a552-edfb1f4b3074">

![image](https://github.com/user-attachments/assets/2d73c54b-92ef-4578-8f97-1ed82b1cc1a4)

**Hardware** 
 - Add-on Release Version: 1.6.0 dev build
 - Blender Version: 4.1.1
- mac m1

**Additional context**
*Scene Links, models, any other helpful context*
the contents shown inside the screenshot is
```
- Collecting pip
  Obtaining dependency information for pip from https://files.pythonhosted.org/packages/d4/55/90db48d85f7689ec6f81c0db0622d704306c5284850383c090e6c7195a5c/pip-24.2-py3-none-any.whl.metadata  
  Downloading pip-24.2-py3-none-any.whl.metadata (3.6 kB)  
  Downloading pip-24.2-py3-none-any.whl (1.8 MB)  
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.8/1.8 MB 2.5 MB/s eta 0:00:00  
  Installing collected packages: pip  
  Attempting uninstall: pip  
    Found existing installation: pip 23.2.1  
    Uninstalling pip-23.2.1:  
      Successfully uninstalled pip-23.2.1  
  Rolling back uninstall of pip  
  Moving to /Applications/Blender.app/Contents/Resources/4.1/python/lib/python3.11/site-packages/pip-23.2.1.dist-info/  
   from /private/var/folders/1w/v1sqb03s6z10r_w866h3_sw00000gp/T/pip-uninstall-4iz5_jkl  
  Moving to /Applications/Blender.app/Contents/Resources/4.1/python/lib/python3.11/site-packages/pip/  
   from /private/var/folders/1w/v1sqb03s6z10r_w866h3_sw00000gp/T/pip-uninstall-3g2fz5lv  
[...]

PR #309, opened 2024-07-31
Fix dependency installation error reporting and handling.

Exairnous:
False positives in the error reporting when installing dependencies caused confusion to users and sometimes led to the dependency not getting installed at all when false errors were detected and the install halted.
Switch to relying solely on is_module_available to determine whether an install was successful and suppress errors unless they are 100% confirmed to be an issue.
Make upgrading pip optional instead of required.
Modify is_module_available to be able to check both system dependencies (i.e. pip) and locally installed dependencies.
Modify is_module_available to explicitly return a bool.

This PR should hopefully address #308 as well as any other similar cases.


Exairnous:
@hrithikwins Can you check and see if this fixes #308 for you, please?

PR #318, opened 2024-11-08
Created tool tips for Add Component Menus and all component types

Spiderguy-F:
For some components, I think I don't have the best explanation. If there is a source for that, or you want me to change a specific tool tip, let me know.


Spiderguy-F:
We solved this differently...

PR #322, opened 2024-12-08
Update glTF add-on imports to support Blender 4.3

Exairnous:
A lot of the files in the glTF add-on were renamed to be shorter to better accommodate the path length limit in Windows for the Blender 4.3 release (https://github.com/KhronosGroup/glTF-Blender-IO/pull/2331).  This prevents the Hubs add-on from loading at all because many of the previous imports are no longer valid.

This PR fixes the imports and restructures the glTF add-on imports so that imports common to all versions are done first, then each Blender version has a section for the rest of it's imports from the glTF add-on. This results in some duplication of imports, but keeps things clear and simple.

Cleanup:
* Several imports from the same file have been combined into a single import.
* Imports at the beginning of the file have been ordered alphabetically (mostly).
* Imports at the beginning of the file with the format `from x import y` have been placed after imports with the format `import x`.
* One of the`BLACK_LIST` imports has been moved to the top of its function to be more readable with its increase in complexity.
* Added a comment to describe why `BLACK_LIST` is imported in a function instead of at the beginning of the file.

### DIFF
--- a/addons/io_hubs_addon/io/gltf_exporter.py
+++ b/addons/io_hubs_addon/io/gltf_exporter.py
@@ -6,8 +6,8 @@ from ..components.utils import get_host_components
 import traceback
 
 if bpy.app.version < (3, 0, 0):
-    from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
     from io_scene_gltf2.blender.exp import gltf2_blender_export
+    from io_scene_gltf2.io.exp.gltf2_io_user_extensions import export_user_extensions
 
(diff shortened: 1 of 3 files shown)


Spiderguy-F:
When I try to install your PR in windows 11, Blender 4.3.2 I get no messages at all, but also none of the functionality. Is that the error you were expecting?

Exairnous:
I'm merging this since there have been no objections and @Spiderguy-F gave his approval in Discord (https://discord.com/channels/498741086295031808/964183314674098276/1320375581698949121).

PR #330, opened 2025-11-12
CI: runs tests on 4.x versions of Blender, drops tests on unsupported versions

DougReeder:
## What?
CI: runs tests on versions 3.6–5.1 of Blender, drops tests on unsupported versions

## Why?
We need to test against all supported versions, but not against unsupported versions

## How to test
1. Navigate to Actions tab of this repo
2. Observe that all tests pass

The docs don't state what versions of Blender this works with, so no changes are necessary

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -57,8 +57,9 @@ jobs:
     needs: lint
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        version: ["3.5.0", "3.4.0", "3.3.1", "2.93.9"]
+        version: ["3.6.23", "4.0.2", "4.1.1", "4.2.19", "4.3.2", "4.4.3", "4.5.8", "5.0.1", "5.1.1" ]
     timeout-minutes: 20
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2


Exairnous:
Before we can drop unsupported versions and add the supported versions we need to decide what versions we want to support (and then we should note it down as our policy).

The initial informal policy for the add-on was to only support the latest Blender release, but I have always thought that was too inflexible and it relaxed as time went by, though no formal policy was ever adopted.

Some possible policies we could adopt (in no particular order).
1. Support the currently maintained LTS versions of Blender, along with the latest version of Blender.
2. Support all the versions from the oldest currently maintained LTS to the latest version of Blender.
3. Support all the versions of a series, i.e. what you have here.
4. Support all the versions from a team decided starting version to the latest version of Blender.
5. Support several versions that are team decided.

UPDATE: Include Doug's option from https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/330#issuecomment-3549113339:

6. Support all the versions from the oldest currently maintained LTS to the latest version of Blender, as well as any additional specific releases that are decided on.

For the options that contain team decided versions, what team should decide is another question.  I would say that probably the art team should decide, but arguments can be made for the programming and governance teams too.

My preference would probably be for option 2 or option 4.  Fun fact, I still predominantly use Blender 3.6 because the EEVEE shaders compile exceptionally quicker than future LTS releases.  I just took some rough numbers and this is how long it takes for the EEVEE shaders to compile on the same, fairly simple, scene for the different versions:

4 seconds for Blender 3.6.14 LTS
4 seconds for Blender 4.0.1
4 seconds for Blender 4.1.0
107 seconds for Blender 4.2.1 LTS
42 seconds for Blender 4.5.1 LTS
42 seconds for Blender 5.0 beta

So even though it would normally make sense, I'm not sure whether I really want to drop support for Blender 3.6 yet.

Another thing to think about is what we do when we drop support for a version, i.e. do we just not test on it anymore, or do we remove any version specific support for it that was introduced?

@jua360 @hobbs-Hobbler @Spiderguy-F @GottfriedHofmann @aelainelong @theanine3D Feel free to weigh in on this as well (and feel free to tag anyone else who you think would be appropriate).

hobbs-Hobbler:
I definitely have opinions on needing the flexibility to jump **between** versions of Blender.

### Versions

In my Tron build, as I advertised, the heavy lifting was done by the **textures**. The essential 3D objects were 3 cubes and 1 cylinder which looked like busy, computer-like buildings and roads because of only 5 textures. Following a tutorial, I built the textures first in Blender 4.4.3 (and GIMP) simply because the drawing tools are more visible and I'm still learning how to draw in Blender (grease pencil) and I don't know how to easily find the same tools in some of the earlier versions.  (Note: the textures, even though I "drew" them, are primarily straight lines, dots, and dashes.  It is only clever artwork that changes them to look like roads, RAM, and motherboard components.)

I managed to make the Blender version of the build quickly and the visuals. I'll attach 2 Work-In-Progress images that matched the tutorial well (props to that teacher!).  But initially what I saw in Blender translated to Hubs ZERO (Read: it was completely black b/c it was, essentially, unlit.)

I've been having terrible alpha sorting errors in Blender 4.4 and I simply cannot reliably build an immersive space (via a glb file) in anything higher than Blender 4.0 right now.  Alpha sorting errors in a space where I was depending on transparency is bad news. Blender tutorials are happy to show you all of the new features of these 4.x versions, but in order to be sure things like lighting, fog, and the feel of the space work, I MUST use the Hubs Add-On to test spaces.  So I went out of my way to open and use Blender 4.0 anytime I was working on the entire space, even though components like the textures, Light Cycle, and Solar Sailer were built in 4.4.

So my first point is that **artists and creators _NEED a reliable version_ to depend on when the stakes are high.**  I'd already promised a Hubs build for Halloween before I moved the glb into Hubs.  Finding a finished and lit scene in Blender go totally black in Hubs will make you realize that there is a LOT more work to do.

### Testing

For the Tron build, I spent easily 2 weeks daily testing small tweaks with the Hubs Add-On. It took 3 days to get the lighting right including considering that Flynn's Hideout has LOTS of silver/reflective textures so I had to decide what I could accept being seen in those reflections (like the dining room table).  

For the larger city part, I had to fight between alpha vertex painting (which I eventually abandoned even though I thought I had a perfect and easy test case) and went with black fog to get the main building light beam to fade into the sky (a VERY important art feature for the Tron feeling that I was not willing to compromise on. I did NOT want the "top" of the beam to be easily seen. The beam could not look "tinker toy", it had to look like a power light beaming into space).  

Further, I worked hard on the glowing tail texture that follows the Light Cycle (it glows from the top _and_ bottom).  So again, I had to hop between GIMP, Blender 4.0, and Blender 4.4 to get the transparency and color drop off just right.  Repeat again since I re-did the same Light Cycle in purple.  Finally, testing with the Hubs Add-on was critical to making sure that I got the art effect of "light propelled/light speed" that I was looking for.

So my second point is that **I could not have made the Tron build (which was finished with Blender 4.0) without the Hubs Add-On**.  Without testing, I would have had a cute few seconds of Blender video zooming down a digital street....and that's it.

Notice that I'm testing "how much" bloom intensity and that the textures in the back are still their original Blender white (b/c Hubs did not understand the blue color ramp).
Tron_WIP1
Tron_WIP2

DougReeder:
I think we should *support* (test on) all the currently supported versions of Blender (including LTS versions).  I don't think we should *remove* code unless it makes non-trivial difficulties.  

So, the plugin should continue to work with old versions of Blender, but we don't put any effort into doing so. If a new version of the plugin loses compatibility with an old version of Blender, users can still use an old version of the plugin.

Of course, we could make a special exception for a particular unsupported version of Blender, if the Art and Programming teams agreed.

theanine3D:
I have successfully been maintaining compatibility with as far back as Blender 3.0 in my own Blender addons. So it's doable - but it does require some mindfulness and extra code to check for versions. Since we have a very limited amount of people, I would say being practical has to be the winning option.

I would go with option 1: 
**Support the currently maintained LTS versions of Blender, along with the latest version of Blender.**

We don't need to outright block older versions - they should still be able to install and try running the addon, and much of the addon will likely still work. But we won't provide support for anything broken if they're running an unsupported version.

DougReeder:
It looks like @Exairnous 's option 2

> Support all the versions from the oldest currently maintained LTS to the latest version of Blender.

...with the possibility of adding specific releases, would satisfy everyone.

That makes the current list:
"4.2.15", "4.3.2", "4.4.3", "4.5.4", "5.0.0"

Exairnous:
Thank you to everyone so far who has provided opinions on the policy for what Blender versions to support.  I want to mull things over a bit before responding further, but in the mean time, I wanted to note a few things I found out about the tests failing in Blender 4.5 and 5.0.

It seems that our support for HDR images may be what's causing the issue.  The `uri` for the image isn't being set when the tests are run, which causes a python error in the glTF add-on (this seems to be the specific line here: https://github.com/KhronosGroup/glTF-Blender-IO/pull/2560/files).  It seems that this is actually a problem with exporting the glTF in separate parts (a `.gltf` file, .a `bin` file, and any images) with the `glTF Separate` format option since the `glTF Binary` and `glTF Embedded` format options don't encounter this error.  If I export the environment-settings test file using either the binary or embedded formats, everything works fine, but if I use the separate format I get the same error the tests encounter (which also export using the separate format).

Since the embedded format is text based, we could potentially switch to using the embedded format for the tests, but this has the downside of sightly larger file sizes and potentially just sweeps the problem under the rug/out of sight.

I'll try to keep this PR updated with anything additional I find out about this issue.

Exairnous:
Turns out there's an additional call needed to set the `uri` in Blender 4.5+ for `glTF Separate`.  I also found there are issues with the test for the Loop Animation component.  I'll put up one or more PRs to fix the tests when I get a chance.

DougReeder:
From additional discussion, it now appears option 4, with the earliest version set to 4.0, would satisfy everyone.  That makes the current CI versions:
"4.0.2", "4.1.1", "4.2.15", "4.3.2", "4.4.3", "4.5.4", "5.0.0"

HussainAther:
I checked this out locally (`gh pr checkout 330`). +1 to dropping EOL Blender versions in CI.

To help this land cleanly, it might be worth stating a simple CI support policy going forward, e.g.:
• test the oldest still-supported LTS
• test the current LTS
• test latest stable (optionally treat next-major as best-effort)

With that framing, testing 4.2–4.5 makes sense. One open question is whether we also want to keep 3.6 LTS in the matrix for now, since it’s still widely used in the community. I’m happy to follow up with a small tweak if the group wants to include or explicitly drop it.

HussainAther:
One small thought: if we keep Blender 5.0.0 in the matrix for early warning, it might be best-effort / allowed-to-fail until it’s fully settled, so CI doesn’t get flaky.

HussainAther:
Added Blender 3.6 LTS (3.6.23) to the CI matrix to cover the oldest still-commonly-used LTS while keeping the 4.x series coverage.

Exairnous:
@HussainAther Thank you for your input.

> One open question is whether we also want to keep 3.6 LTS in the matrix for now, since it’s still widely used in the community.

That is a good question.  I still predominantly use 3.6 LTS, but I'm not aware how widely it is used in the community.  Can you provide some numbers/additional details to back up this claim?

HussainAther:
My thinking was mainly based on:

- Blender 3.6 being the current LTS release, which typically has a longer
adoption tail in production pipelines
- Many tutorials, assets, and existing projects still targeting 3.6 LTS for
stability
- Personal experience still seeing 3.6 used frequently in XR / WebXR
workflows

That said, I don’t have hard usage metrics to quantify how widespread it
currently is.

If the goal is to keep CI lean and focused on actively supported versions,
I’m totally fine with dropping 3.6 - I just wanted to raise the question in
case maintaining LTS coverage was still considered valuable.

Happy to defer to maintainers on the direction here.


Syed Hussain Ather (he/him)
<https://github.com/hussainather>
<https://scholar.google.com/citations?user=jPL2-dEAAAAJ&hl=en>
<https://www.linkedin.com/in/syed-hussain-ather-049919137/>
<https://twitter.com/shussainather>
http://hussainather.com

Exairnous:
@HussainAther Thank you for the elaboration.  This is good to know.

> Blender 3.6 being the current LTS release

Actually, the currently supported LTS releases are Blender 4.2 and Blender 4.5 (though in another few months it's going to be 4.5 and 5.2).  That being said, I never liked Blender 4.2, so I wouldn't mind keeping support for Blender 3.6 (and it's only one version earlier than the earliest version we'd been thinking of supporting anyway).

@DougReeder Since I'm predominantly maintaining the add-on right now, let's add 3.6 to the list of supported versions, so it would be 3.6 - 5.1.

On another note, while I still think for now we should continue with option 4, I'm starting to wonder if in the future we might want to think about switching to your option (which I've added to my comment as option 6, so they're all together) because there may be some functionality we want that is only present in older blender versions (e.g. I thought [backface culling](https://projects.blender.org/blender/blender-manual/issues/105059) might be one, but it turns out it was just a documentation issue) and we may not want to have an unbroken chain of supported versions back to them (not sure). 

Anyway, to summarize, let's go with option 4 for now and support Blender versions 3.6 - current (5.1), but potentially re-evaluate this policy at a later date as we see how things unfold.

Exairnous (approved):
LGTM.  Thanks.  Merging.

PR #333, opened 2026-03-17
Fix migration when enabling the add-on in Blender 5.1

Exairnous:
## What?

Explicitly specifies the execution context when calling `migrate_hubs_components` during the registration of the add-on in the middle of a session instead of relying on the execution context being passed automatically.

## Why?

Blender 5.1 now requires the execution context to be specified as well if you want to push an undo step when calling an operator.  Blender 5.1 no longer auto-detects which option you specified if you only specify one and assumes that you are specifying the execution context.  If you try to only specify whether to push an undo step without specifying the execution context, it fails with the following error:

```Python
Traceback (most recent call last):
    File "/path/to/io_hubs_addon/__init__.py", line 67, in registration_migration
    bpy.ops.wm.migrate_hubs_components(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        True, is_registration=True)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: _bpy.ops.call() argument 3 must be str, not bool
```

## Examples

N/A

## How to test

1. Open Blender 5.1 from a terminal
2. Disable the add-on (if enabled)
3. Enable the add-on.
4. See that no error about `bpy.ops.wm.migrate_hubs_components` is printed to the terminal.
5. If you have a blend file with components from an add-on before version 1.0, load that with the add-on disabled, enable the add-on, and see that the migration happens as expected.
6. If you repeat the above without this fix, you should see an error about `bpy.ops.wm.migrate_hubs_components` printed to the terminal when enabling the add-on.

No functionality is changed, this just ensures the current behaviour is maintained for Blender 5.1.

## Limitations

None.

None.

## Open questions

None.

The change in behaviour is potentially because of the Python upgrade from 3.11 to 3.13, but this specific change (with regard to calling operators) is unreported in the 5.1 release notes.

The specified execution context is the default execution context for operators (so it should be the same as what was automatically passed previously).

Previous versions of Blender shouldn't be affected because both positional options have always been present in this order.

Blender 3.6 and previous also had an option to override the context as the first positional argument (which isn't specified here), but not specifying this shouldn't be a problem as it should just detect which options we're passing.

### DIFF
--- a/addons/io_hubs_addon/__init__.py
+++ b/addons/io_hubs_addon/__init__.py
@@ -65,7 +65,7 @@ def register():
         def registration_migration():
             # Passing True as the first argument of the operator forces an undo step to be created.
             bpy.ops.wm.migrate_hubs_components(
-                True, is_registration=True)
+                'EXEC_DEFAULT', True, is_registration=True)
         bpy.app.timers.register(registration_migration)
 
 


Exairnous:
Update: it now looks like the issue was introduced in https://projects.blender.org/blender/blender/pulls/145344.  I have filed a bug report with Blender: https://projects.blender.org/blender/blender/issues/155928

GottfriedHofmann:
> Update: it now looks like the issue was introduced in https://projects.blender.org/blender/blender/pulls/145344. I have filed a bug report with Blender: https://projects.blender.org/blender/blender/issues/155928

And already fixed in master: https://projects.blender.org/blender/blender/commit/9afe6f3b519c6ef1edd3aa68d67222b119db2e76

Exairnous:
Yeah, they did that fast. They've also added it as a [candidate to backport to 5.1.1](https://projects.blender.org/blender/blender/issues/155533) :D


I still think this PR should be applied, though, because it won't hurt anything and it will safeguard 5.1.0 in case anyone doesn't update.

Exairnous:
As discussed in the 2026-03-24 Hubs Dev Meetup ([notes link](https://docs.google.com/document/d/1OMh3-_DpjHABMMrXtECzLPNxipQMixx32l6K4zqSoYY/edit?tab=t.0#heading=h.qijp3clx090)), since I haven't been able to get a review on this, I am merging it on my own recognizance.

PR #334, opened 2026-04-04
Add pre-export validation warnings to scene debugger export flow

HussainAther:
## What?
Adds a lightweight pre-export validation step to the Blender scene debugger export flow.

Before calling `bpy.ops.export_scene.gltf(...)`, the add-on now checks the scene for common issues and reports warnings, including:
- mesh objects with no assigned material
- very high polygon meshes
- unapplied object scale
- materials with image texture nodes that have no image assigned
- hidden or non-renderable objects that may still export depending on export settings

Additionally, initializes `viewpoint` before the `try` block in `HubsUpdateRoomOperator.execute` to avoid potential reference-before-assignment issues in error cases.

## Why?
Currently, the scene debugger export path does not perform any validation before export. This can lead to:
- confusing export failures
- unexpected behavior in Hubs after export
- difficulty debugging scene issues

This PR introduces a minimal validation layer to improve user feedback and debugging without changing the existing export workflow or blocking valid exports.

## Examples
Console output example before export:

### DIFF
--- a/addons/io_hubs_addon/debugger.py
+++ b/addons/io_hubs_addon/debugger.py
@@ -13,7 +13,110 @@ DEMO_SERVER_URL = ""
 ROOM_FLAGS_DOC_URL = "https://github.com/Hubs-Foundation/hubs-docs/blob/master/docs/hubs-query-string-parameters.md"
 
 
+def _is_identity_scale(scale, eps=1e-6):
+    return (
+        abs(scale[0] - 1.0) < eps and
+        abs(scale[1] - 1.0) < eps and
+        abs(scale[2] - 1.0) < eps
+    )
+
+
+def _has_missing_texture_nodes(material):
+    if not material or not material.use_nodes or not material.node_tree:
+        return False
+
+    for node in material.node_tree.nodes:
+        if node.type == 'TEX_IMAGE' and not getattr(node, "image", None):
+            return True
+
+    return False
+
+
+def validate_scene(context):
+    """
+    Conservative pre-export validation.
+
+    Errors:
+      - none currently, but this is the right place to add hard blockers later
+
+    Warnings:
+      - mesh has no material
+      - mesh is very high poly
+      - mesh has unapplied scale
+      - material contains image texture nodes with no image
+      - hidden objects may still export depending on export prefs
+    """
+    errors = []
+    warnings = []
+
+    export_prefs = context.scene.hubs_scene_debugger_room_export_prefs
+
+    for obj in context.scene.objects:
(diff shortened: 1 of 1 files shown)


Exairnous (changes requested):
Thank you for the PR.

I've only given this a visual review, but the general idea seems okay, adding more validation to the exporter has been discussed in the past and is something we want.  But, this only affects the scene debugger and introduces an additional loop over everything in the scene, which would slow down the export.  Instead, the validations should be added to the appropriate hook methods in `class glTF2ExportUserExtension` in https://github.com/Hubs-Foundation/hubs-blender-exporter/blob/master/addons/io_hubs_addon/io/gltf_exporter.py

For a UI display to the user, the `hubs_report_viewer` should be used and an `export_report` should be generated (similar to what's done in https://github.com/Hubs-Foundation/hubs-blender-exporter/blob/master/addons/io_hubs_addon/io/gltf_importer.py for the `import_report`), but this would be a much more involved change, so it's up to you whether you would want to try to include a UI display in this PR.  It also would be slightly different from what's done with `gather_import` as we don't call the hooks directly (they're called by the glTF exporter, so we can't pass extra arguments in), so you'd just declare the export_report global in the hook and append to it.

Also, there seem to be a lot of formatting changes in this.  If this is the result of running `format.py` and is actually required, then please submit that as a separate PR, otherwise, please don't include vast formatting changes along with features/bug fixes.

Additionally, given the size of the PR, it will probably need to wait until after the sprints for further review, but that doesn't mean we don't want it, and a few of your other recent PRs may be small enough to be included in the upcoming release.

HussainAther:
ty  for the review and for flagging this.

AI tools were indeed used to assist with drafting and structuring parts of this change. That said ,the  code and behavior were reviewed, tested, and validated by me to ensure consistency with the exporter architecture and expected functionality.

I'm ,v h appy to adjust anything if needed to better align with project the integrity of the project. thanks.

Exairnous:
> ty for the review and for flagging this.

You're welcome :slightly_smiling_face: 

> AI tools were indeed used to assist with drafting and structuring parts of this change. That said ,the code and behavior were reviewed, tested, and validated by me to ensure consistency with the exporter architecture and expected functionality.
> 
> I'm ,v h appy to adjust anything if needed to better align with project the integrity of the project. thanks.

As long as you make any changes needed to adhere to the relevant rules and disclosure guidelines specified in https://github.com/Hubs-Foundation/policies-procedures-guidelines-public/pull/17, you should be good.  Probably, most of what needs to be done is disclosure related now.

PR #338, opened 2026-04-06
Ensure video validation is on this branch

HussainAther:
## What?
Adds a pre-export validation warning to the Video component.

If the Video component is missing its required source URL (`src`), it now returns a warning from `pre_export()`.

## Why?
Currently, component configurations are not validated prior to export, which can lead to missing or broken media in Hubs scenes.

This PR adds a small, component-specific validation using the existing `HubsComponent.pre_export()` hook.

## Examples
Example warning:

`Cube: Video component missing source URL`

## How to test

2. Add an object to the scene
3. Add the Video component to the object
4. Clear the "Video URL" field
5. Trigger export (e.g., Update Room Scene or Publish Scene)
6. Check the console for the warning

No documentation changes required.

## Limitations
- Only Video component validation is included in this PR
- Warnings are currently only printed via console

- Centralized validation was considered but deferred in favor of component-level validation

## Open questions
None

Part of incremental work toward broader component-level validation.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/video.py
+++ b/addons/io_hubs_addon/components/definitions/video.py
@@ -19,6 +19,15 @@ class Video(HubsComponent):
         'version': (1, 0, 0)
     }
 
+    def pre_export(self, export_settings, host, ob=None):
+        warnings = []
+
+        # Validate that a video source is provided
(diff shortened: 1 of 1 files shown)


Exairnous:
Thank you for the PR.

PR #340, opened 2026-04-06
Feature/interaction metadata component

HussainAther:
## Summary
Adds a new `interaction-metadata` Hubs component for exporting lightweight interaction/system metadata from Blender objects and bones.

## Motivation
This provides a simple way to annotate exported assets with structured metadata that can be consumed by downstream interaction systems, behavior graphs, or simulation-aware tooling.

## Exported fields
- `behaviorType`
- `interactionProfile`
- `simulationTag`

## Notes
- Follows the existing component-based exporter architecture
- Supports object and bone panels
- Exports only non-empty fields

### DIFF
--- a/addons/io_hubs_addon/components/definitions/interaction_metadata.py
+++ b/addons/io_hubs_addon/components/definitions/interaction_metadata.py
@@ -1,0 +1,47 @@
+from bpy.props import StringProperty
+from ..hubs_component import HubsComponent
+from ..types import Category, PanelType, NodeType
+
+
+class InteractionMetadata(HubsComponent):
+    _definition = {
+        'name': 'interaction-metadata',
+        'display_name': 'Interaction Metadata',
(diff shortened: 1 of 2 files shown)


Exairnous:
Thank you for the PR.

I've only given this a visual review, but it looks like it adds a component that isn't currently supported by Hubs.  Also, Behavior Graphs has it's own add-on and expects very different data to be exported.  So unless you are intending to open a companion PR on the Hubs client (and it is accepted), I think this will likely be closed.

Please note that the add-on allows you to load custom components from user specified directories in its preferences, so you don't have to get a component added to the main add-on in order to use it.  See https://github.com/Hubs-Foundation/hubs-postprocessing-addon as an example (see the readme for the hubs-postprocessing-addon for installation instructions).

Additionally, your PR description doesn't follow the PR template and omits some required fields.

PR #344, opened 2026-04-13
Fix exporting glTF Separate with images

Exairnous:
## What?

Adds a call to the glTF add-on's set_real_uri function when gathering images for export in the glTF Separate format and adds a comment for where to reference future changes that may be needed.

## Why?

To fix the failing environment-settings, lightmap, and reflection-probe tests reported in https://github.com/Hubs-Foundation/hubs-blender-exporter/issues/331.

The comment was added so that next time this function needs updating we'll have a starting point for how to update it.

## Examples

N/A

## How to test

1. Open https://github.com/Hubs-Foundation/hubs-blender-exporter/blob/master/tests/scenes/environment-settings.blend in Blender 4.5+ (or any other blend file that has a component that uses an HDR image)
2. Export a glTF using the glTF Separate format.
3. See that the export completes successfully and that the images have been exported.
4. Run the tests on Blender 4.5 and see that they all pass.

No functionality has changed, so no documentation update is needed.

## Known omissions or limitations

The issue fixed here is only part of why the tests are failing in recent Blender versions, so this won't make all the tests pass in Blender 5.0+.  See https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/330 for more details on the issues with the tests.

None.

## Open questions

None.

The set_real_uri function, and it's use when exporting images in glTF Separate, was introduced in https://github.com/KhronosGroup/glTF-Blender-IO/pull/2515

Related to: https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/345

### DIFF
--- a/addons/io_hubs_addon/io/utils.py
+++ b/addons/io_hubs_addon/io/utils.py
@@ -7,7 +7,20 @@ from io_scene_gltf2.io.com import gltf2_io
 from typing import Optional, Tuple, Union
 
 # Version-specific imports
-if bpy.app.version >= (4, 3, 0):
+if bpy.app.version >= (4, 5, 0):
+    from io_scene_gltf2.blender.com import extras as gltf2_blender_extras
+    from io_scene_gltf2.blender.exp import joints as gltf2_blender_gather_joints
+    from io_scene_gltf2.blender.exp import nodes as gltf2_blender_gather_nodes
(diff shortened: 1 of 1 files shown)


hrithikwins (approved):
LGTM

This works with blender 5.0.0 and it exports the images and hdr separately

PR #345, opened 2026-04-13
Fix Loop Animation component export and tests in recent Blender versions

Exairnous:
## What?

* Fixes the Loop Animation component export with the new Actions API introduced in Blender 4.4 and the tests that are failing in Blender 5.0+.
* Adds new validation checks for the new Actions API and cleans the checks up a bit.
* Improves the tests to check the names exported to the component with the actual names of the animations that were exported and so better detect naming convention changes.  This is also required for the tests to pass in versions prior to Blender 4.4.

## Why?

The tests for the Loop Animation were failing in Blender 5.0+ due to API removals in Blender 5.0 and the actual export isn't correct in some situations since Blender 4.4 (despite the tests passing - they only checked for whether the export matched what was expected, not whether it would actually work in Hubs).  Not everything was properly validated in the UI of the Loop Animation component for versions above Blender 4.4 either.

## Examples

N/A

## How to test

1. Open https://github.com/Hubs-Foundation/hubs-blender-exporter/blob/master/tests/scenes/loop-animation.blend in Blender 4.2, 4.4, and 5.0 (and any others you want to test).
2. Export a glTF using the glTF Separate format for each Blender version.
3. Open the exported `.gltf` file for each Blender version in a text editor and see that the names in the clip property of the Loop Animation component match the names of the actual animations.
4. Run the tests on Blender 4.2, 4.4, and 5.0, and see that the Loop Animation component tests pass.
5. To check the new verification messages:
  - Add an action without a slot and select it in the Loop Animation component, then click on the track in the Loop Animation component to view the warning.  Repeat but push it down to an NLA track before selecting it in the loop animation component (you may have to remove the slot from the action again after pushing it down).
  - Push an action down to an NLA track, then unlink the action slot from the NLA track in the `Action Clip` subpanel in the N-Panel in the NLA track editor (you will need to select the strip for that track for it to show in the N-Panel), then select the NLA track in the Loop Animation component, then click on the track in the Loop Animation component to view the warning.
  - Add an action, and a slot but don't add any keyframes, and select it in the Loop Animation component, then click on the track in the Loop Animation component to view the warning for there being no layer in the action.  Repeat but push it down to an NLA track before selecting it in the loop animation component.
  - Add an action, and a slot, then add some keyframes and delete them, and select it in the Loop Animation component, then click on the track in the Loop Animation component to view the warning for there being no animation present in active slot for the action.  Repeat but push it down to an NLA track before selecting it in the loop animation component.
  - Add an action, and a slot, then add some keyframes, then unlink the slot and add a new slot, then select the action in the Loop Animation component, then click on the track in the Loop Animation component to view the warning for there being no animation present in the active slot for the action.  Repeat but push it down to an NLA track before selecting it in the loop animation component.
  - The warning about "No strips are present in the action layer" is to prepare for future versions of Blender and I don't think there's a way to trigger it currently.

Mostly this just fixes bugs in recent Blender versions and doesn't change any functionality, so no documentation update is needed.  The changes to the validation include warnings to the user and so serve as its own documentation (these can be found in `addons/io_hubs_addon/components/definitions/loop_animation.py`).

## Known omissions or limitations

The issue fixed here is only part of why the tests are failing in recent Blender versions, so this won't make all the tests pass in Blender 5.0+.  See https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/330 for more details on the issues with the tests.

None.

## Open questions

None.

Release notes for Blender 4.4 where the new Action API is introduced:
https://developer.blender.org/docs/release_notes/4.4/python_api/#slotted-actions

Release notes for Blender 5.0 where the legacy Action API is removed:
https://developer.blender.org/docs/release_notes/5.0/python_api/#animation-rigging

Related to: https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/344

### DIFF
--- a/addons/io_hubs_addon/components/definitions/loop_animation.py
+++ b/addons/io_hubs_addon/components/definitions/loop_animation.py
@@ -398,12 +398,16 @@ def is_valid_shape_key_track(ob, track):
     return False
 
 
-def get_animation_name(ob, track):
+def get_animation_name(ob, track, export_settings):
     if is_valid_regular_action(ob, track) or is_valid_shape_key_action(ob, track):
         return track.name
     elif is_valid_regular_nla_track(ob, track) or is_valid_shape_key_nla_track(ob, track):
-        return track.track_name if not is_default_name(
-            track.track_name) else track.action_name
+        if bpy.app.version >= (4, 4, 0) and export_settings['gltf_animation_mode'] == 'ACTIONS':
+            nla_track = ob.animation_data.nla_tracks.get(track.track_name)
+            return nla_track.strips[0].action.name
+        else:
+            return track.track_name if not is_default_name(
+                track.track_name) else track.action_name
     else:
         return track.name
 
@@ -865,7 +869,7 @@ class LoopAnimation(HubsComponent):
     def gather(self, export_settings, object):
         final_track_names = []
         for track in object.hubs_component_loop_animation.tracks_list.values():
-            final_track_names.append(get_animation_name(object, track))
+            final_track_names.append(get_animation_name(object, track, export_settings))
 
         fps = bpy.context.scene.render.fps / bpy.context.scene.render.fps_base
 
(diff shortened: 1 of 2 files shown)


hrithikwins (approved):
LGTM

Checked it works on Blender 5.0.0

Great work!

PR #352, opened 2026-07-04
Fix Loop Animation shape key track validation

Exairnous:
## What?

Prevents a crash if there aren't any regular actions present when checking if shape key actions have NLA tracks and refactors `is_usable_action` to be more robust, handle shape key actions, and prevents the action not having fcurves potentially being reported as the action not having an active slot.

## Why?

These issues prevent shape key action tracks from being selected in the loop animation component and or report false errors.

## Examples

N/A

## How to test

1. Add the basis shape key and a regular shape key to the default cube.
2. Insert a keyframe on the value of the regular shape key.
3. Add a Loop Animation component.
4. Add a track to the Loop Animation component and select the shape key action.
5. See that the shape key action is added as a track and no errors reported.

Note: attempting this process without this PR would result in no items being shown when attempting to add a track unless a regular action is present and even then attempting to select the shape key action wouldn't add the track properly (it would show up as a blank track).  All of this would result in various Python errors being printed to the terminal.

No functionality is changed, so no documentation update is needed.

## Known limitations

This doesn't add any tests because our current testing framework only tests exports and this is all UI stuff.

None.

## Open questions

None.

None.

### DIFF
--- a/addons/io_hubs_addon/components/definitions/loop_animation.py
+++ b/addons/io_hubs_addon/components/definitions/loop_animation.py
@@ -196,10 +196,11 @@ def has_action(tracks_list, action, invalid_action=None):
 
 def action_has_nla_track(ob, action):
     nla_tracks_with_action = []
-    for nla_track in ob.animation_data.nla_tracks:
-        track_action_name = get_action_name(nla_track)
-        if track_action_name == action.name:
-            nla_tracks_with_action.append(nla_track.name)
+    if ob.animation_data:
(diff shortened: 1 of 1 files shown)


Exairnous:
ROADMAP STATUS UPDATE: This pull request has been added to the short term section of the [Programming Team Roadmap](https://docs.google.com/document/d/1xRSHEgf4jE4SRsx2Z5BX_uXKbJ7KuCW3v0ntugA76io/edit?tab=t.0).

PR #353, opened 2026-07-04
Fix import/export settings panels in Blender 4.4+

Exairnous:
## What?

Converts the automatically called import/export draw function for Blender 4.2+ into two functions (one for import and one for export) and registers them specifically with the glTF add-on.

## Why?

The glTF add-on changed how it handled registering the export panel in Blender 4.4+ in order to better accommodate collection exporters, so this accounts for the changes while still supporting previous versions and also enables the settings panels to be shown in the collection exporters.

## Examples

N/A

## How to test

1. Open Blender.
2. Select File > Import > glTF 2.0 (.glb/.gltf).
3. See that there is a panel for Hubs Components, then expand the panel and see that it declares there are no options yet. 
4. Cancel the import.
5. Select File > Export > glTF 2.0 (.glb/.gltf).
6. See that there is a panel for Hubs Components, then expand the panel and see that it declares there are no options yet. 
7. Cancel the export.

For versions of Blender that support collection exporters:

1. Go to the collection tab in the properties editor.
2. Add a glTF 2.0 exporter.
3. See that there is a panel for Hubs Components, then expand the panel and see that it declares there are no options yet. 

## Documentation of functionality

Mostly no functionality has changed, so no documentation needs to be updated.  Technically, this adds the export settings to the collection exporters, but the functionality is the same as with the regular exporter and there isn't any documentation that I know of that makes sense to update (if there is some that someone else knows of, feel free to mention it and I'll update it).

## Known limitations

None.

## Alternative implementations considered

* Keeping a single, combined function for both import and export panels, but this wasn't as clear and would have been slightly more code.
* Keeping the import panel being called automatically by the glTF add-on, but it was simpler/more consistent to do both panels the same way and hopefully helps future proof everything. 
* Using one or more of the automatically called function names (draw, draw_export, draw_import) for previous versions was considered, but this would have made it more complicated and could have potentially resulted in duplicate panels.

## Open questions

None.

## Additional details or related context

Related glTF commits/PRs:
https://github.com/KhronosGroup/glTF-Blender-IO/commit/fae512b1981493794ccafb4b187f9ada3c5d3b1f

https://github.com/KhronosGroup/glTF-Blender-IO/pull/2313

https://github.com/KhronosGroup/glTF-Blender-IO/commit/49b56804049c4724c89f6be49ea8382c0b63f9da

UPDATE: also fixes https://github.com/Hubs-Foundation/hubs-blender-exporter/issues/347

PR #354, opened 2026-07-04
Add "Original Generator" export setting

Exairnous:
## What?

Adds an export setting that allows you to specify the original generator of a GLB file.

## Why?

So that the provenance of the GLB file can be maintained when it is modified.

## Examples

Before:

After:

## How to test

1. Open Blender.
2. Select File > Export > glTF 2.0 (.glb/.gltf).
3. Expand the Hubs Components panel and see that there is an "Original Exporter" field. 
4. Enter something into the "Original Exporter" field and export a GLB.
5. Open the exported GLB in a text editor and see that the "generator" property has what you entered in the "Original Generator" field, followed by the statement that it was re-exported by the glTF add-on.

For versions of Blender that support collection exporters:

1. Go to the collection tab in the properties editor.
2. Add a glTF 2.0 exporter.
3. Expand the Hubs Components panel and see that there is an "Original Exporter" field. 
4. Enter something into the "Original Exporter" field and export a GLB.
5. Open the exported GLB in a text editor and see that the "generator" property has what you entered in the "Original Generator" field, followed by the statement that it was re-exported by the glTF add-on.

If adopted, this will probably need documentation; however, since it is a very simple feature and at the moment it is mainly for use by a few people (that I will be in contact with) for the branding removal project, I'm not going to bother adding documentation at present.  If we decide we want to include this in the main add-on, I will add the documentation before merging.

## Known limitations

The "Original Generator" property is stored on the scene and shared by all the different ways to export, so all exports within the scene (including any collection exporters) will share the one value.  Having separate values for each export location isn't supported as it isn't needed for the branding removal project.  If separate values for each export location are desired, they can be added later (either in this PR or a separate PR).

None.

## Open questions

* Is this useful for more than the branding removal project and do we want to include this in the main add-on?
* Should separate values for each export location be supported?

This is mainly for the [branding removal project](https://docs.google.com/document/d/1bp2GInoMMae4q00gcxAAjmxVtL2hus-WO-YHhppeuvg/edit?tab=t.0), but may also be useful in general.

Note: this PR is based off of https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/353  After that PR is merged I will update this PR to remove the extraneous commit from https://github.com/Hubs-Foundation/hubs-blender-exporter/pull/353

## Installable zip of this PR
[io_hubs_addon_1.8.0.596.zip](https://github.com/user-attachments/files/30112000/io_hubs_addon_1.8.0.596.zip)

### DIFF
--- a/addons/io_hubs_addon/__init__.py
+++ b/addons/io_hubs_addon/__init__.py
@@ -25,28 +25,33 @@ bl_info = {
 create_prefs_dir()
 
 
-# Blender 4.2+ glTF Extension Import/Export Settings Panel
-def draw(context, layout):
-    layout_header, layout_body = layout.panel('HBA_PT_Import_Export_Panel', default_closed=True)
-    sfile = context.space_data
-    operator = sfile.active_operator
+# Blender 4.2+ glTF Extension Import/Export Settings Panels
+def draw_import_settings(context, layout):
+    layout_header, layout_body = layout.panel('HBA_PT_Import_Panel', default_closed=True)
 
     # Panel Header
-    if operator.bl_idname == "EXPORT_SCENE_OT_gltf":
-        props = bpy.context.scene.HubsComponentsExtensionProperties
-    elif operator.bl_idname == "IMPORT_SCENE_OT_gltf":
-        props = bpy.context.scene.HubsComponentsExtensionImportProperties
-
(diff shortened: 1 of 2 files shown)


Exairnous:
Added a zip of this PR that can be installed to the PR description.

PR #355, opened 2026-07-17
Initial blender 5.2 support

Exairnous:
## What?

Fixes an issue with exporting material components, the test for the frustum component, adds 5.2 to the Blender versions that are tested by CI, and updates some of the previous Blender versions that are tested to their latest bug fix releases.

## Why?

The glTF add-on that the Hubs add-on depends on has changed what they pass to the material hooks and how exported things are sorted (they're sorted by name now) for Blender 5.2.  The CI test updates are so that Blender 5.2 gets included in the testing and because it's theoretically good maintenance to keep everything at the latest released version.

See the commits for more details on the individual changes.

## Examples

N/A

## How to test

For the material fixes:
1. Open Blender from a terminal.
2. Export a scene with a set of connected video texture source and video texture target components in it.
3. See that there are no errors reported in the terminal.
4. Load the exported GLB into Hubs and see that the video texture source/target works as expected.

For the test fixes/updates:
6. Run the tests and see that they all pass.
7. Run the publish workflow and see it completes successfully.

No functionality was changed, so no documentation update is needed.

## Known limitations

None.

Having all Blender versions look up the material by name, but this could theoretically decrease performance for previous Blender versions.

Changing the test for frustum components to look up the object by name instead of by index, but there's no reason for the additional objects to be present in the test file, so it was easier and more consistent to remove the extraneous objects from the test file.

## Open questions

None.

glTF add-on PRs that are related to the fixes here:
https://projects.blender.org/blender/blender/pulls/154741
https://projects.blender.org/blender/blender/pulls/157106


addons/io_hubs_addon/io/gltf_exporter.py
```diff
@@ -165,6 +165,11 @@ def gather_material_hook(self, gltf2_object, blender_material, export_settings):
         if not self.properties.enabled:
             return
 
+        if bpy.app.version >= (5, 2, 0):
+            # The glTF add-on in Blender 5.2+ no longer passes the actual Blender material, so we need to look it up ourselves
+            blender_material = bpy.data.materials[gltf2_object.name]
+
+        blender_material = bpy.data.materials[gltf2_object.name]
         self.export_hubs_components(
```
DougReeder, line 172: Is blender_material set twice on 5.2 and later?

Exairnous: It is, but it shouldn't be.  Good catch!

nurse-the-code:
We tested this locally on my machine on Friday and everything worked. I did not run the automated tests locally, though. Overall, this looks pretty good, although I must admit that I am not very familiar with the code base.

DougReeder (approved):
The code looks ok to me

